PR #238, opened 2026-03-31
Feat/business personal accounts

cmazwar:
## Summary

Add profile-based account classification for bank accounts and transactions. Users can assign accounts to a **Personal** or **Business** profile in Manage Accounts; transactions inherit their account's profile by default. Individual transactions can override this via the transaction detail screen. A global profile filter (All / Personal / Business) is available on both the home screen and the all-transactions screen.

Internally this replaces the binary `isBusiness: Boolean` toggle with a `profileId: Long?` foreign key referencing a new `profiles` table (seeded with Personal id=1 and Business id=2), laying the groundwork for user-defined named workspaces in a future release.

## Type of change

- [x] feat: New feature
- [ ] fix: Bug fix
- [x] docs: Documentation update
- [x] test: Add/update tests

## How to test

1. Go to **Manage Accounts** → tap any account → assign it to **Business** or **Personal** and verify the badge updates
2. On the **Home screen**, tap the profile filter in the top bar → select **Business / Personal / All** and verify transactions filter correctly
3. Scroll up to confirm the filter also appears in the expanded greeting card header
4. Go to **All Transactions** → expand filters → use the profile chip row (**All / Personal / Business**) to filter
5. Open a **Transaction Detail** → verify "Classification" shows the inherited profile (Personal or Business) in read-only view
6. Tap edit on a transaction → use the **Personal / Business** segmented button to override the account default → save and verify the override persists
7. Verify a transaction with an explicit override is filtered by its own profile, not the account's
8. Fresh-install (or clear app data) and re-open → confirm the two default profiles (Personal, Business) are seeded correctly
9. Upgrade from a v36 build → confirm AutoMigration runs without data loss

## Breaking changes

- [ ] No breaking changes
- [x] Yes (describe): Database schema bumped from **v36 → v39** via three additive AutoMigrations (profiles table, profile_id on account_balances, profile_id on transactions). Users on v36 auto-migrate with no data loss. `isBusiness`-based APIs (setAccountBusiness, filterTransactionsByBusiness, BusinessFilter enum) have been removed.

## Checklist

- [x] Tests added/updated (if applicable)
- [x] Docs updated (if applicable)
- [x] `./gradlew test` passes locally
- [x] `./gradlew lint` passes locally


sarim2000 (commented):
## Review: Business/Personal Accounts Feature

### Overall
Solid first pass — the tagging and filtering approach is pragmatic. A few issues to address before merging.

---

### Issues

**1. Unrelated code in this PR — `CategoryBudgetLimitEntity` / `CategoryBudgetLimitDao`**
These files (`CategoryBudgetLimitEntity.kt`, `CategoryBudgetLimitDao.kt`) and their registration in `PennyWiseDatabase.kt` have nothing to do with business/personal accounts. They should be in a separate PR. Including them in the v36→v37 migration ties an unrelated schema change to this feature, making rollback harder.

**2. `HomeViewModel` fetches 50 recent transactions just to take 3**
```kotlin
transactionRepository.getRecentTransactions(limit = 50),
```
You're loading 50 transactions to filter by business and then `.take(3)`. This is wasteful — consider either:
- Doing the business filter at the DB query level, or
- At minimum, documenting why this over-fetch is necessary

**3. Duplicated business filter logic in `HomeViewModel`**
The account-level business filtering is copy-pasted 3 times:
```kotlin
when (businessFilter) {
    BusinessFilter.ALL -> true
    BusinessFilter.PERSONAL -> !account.isBusiness
    BusinessFilter.BUSINESS -> account.isBusiness
}
```
Extract this into a helper (e.g., `filterAccountsByBusiness()` alongside the existing `filterTransactionsByBusiness()`).

**4. Business filter dropdown appears in TWO places on HomeScreen**
The filter is in both the top app bar (circle icon button) AND inside the `GreetingCard`. This means the same control is rendered twice. Pick one location — probably the top bar icon since the GreetingCard is the cover-style variant.

**5. Transaction detail business toggle logic is confusing**
```kotlin
onClick = { viewModel.updateBusinessStatus(if (isAccountBusiness) false else null) }  // Personal
onClick = { viewModel.updateBusinessStatus(if (!isAccountBusiness) true else null) }   // Business
```
The null-vs-explicit logic here is hard to follow. The intent is "set to null if it matches the account default, otherwise set explicitly" — but there's no explanation. A comment or a named helper would help. Also, this means toggling Personal→Business→Personal doesn't round-trip cleanly if the account default changes later.

**6. `TransactionsViewModel` — business filter state is duplicated**
The filter lives in both `_businessFilter` (local MutableStateFlow) AND `userPreferencesRepository.businessFilter`. The local state is set from the preference flow, but `setBusinessFilter()` writes to preferences and waits for it to propagate back. This works but is fragile — if anyone reads `_businessFilter` before the preference write propagates, they get stale state.

**7. Missing `is_business` default in `TransactionEntity`**
```kotlin
val isBusiness: Boolean? = null
```
No `defaultValue` annotation on the `@ColumnInfo`. The auto-migration from 36→37 will add the column — make sure the default in the schema JSON matches. Room should handle `null` for nullable columns, but worth verifying the exported schema.

**8. `cachedAccountBalances` used in `HomeViewModel.getBusinessAccountKeys()` but cache timing is unclear**
`getBusinessAccountKeys()` reads from `cachedAccountBalances`, but it's not obvious when this cache is populated vs when the filter functions are called. If `filterTransactions` runs before the balances are loaded, business filtering silently fails (all transactions treated as personal).

---

### Minor / Nits

- `Filters.kt` — file lost its trailing newline
- `AccountBalanceDao.kt` — whitespace-only changes (trailing spaces removed) mixed with real changes. Not a blocker but makes the diff noisy.
- `GreetingCard.kt` — the `remember(today)` change for `subtitle` is a bugfix unrelated to this feature. Fine to include but worth noting.
- `TransactionItem.kt` — the "Business" tag in subtitle only shows for `isBusiness == true`, not when the account is business and transaction inherits. Inconsistent with how filtering works.

---

### What's good
- Clean separation: `filterTransactionsByBusiness()` as a shared utility
- Account-level default with transaction-level override is a sensible model
- Filter persisted in DataStore so it survives restarts
- The ManageAccounts context menu toggle is intuitive

### Suggested before merge
1. Remove `CategoryBudgetLimit*` from this PR
2. Deduplicate the account filtering logic
3. Remove one of the two filter dropdown locations on Home
4. Verify the schema export for the nullable `is_business` column on transactions

cmazwar:
either u reviewed without last commit or i am missing something. isBusiness and related work was previous pr and since has been redoen with profileid filtering. also from your comments @sarim2000 last commit already fixed quite some.

Already resolved :
- Issue 1 (CategoryBudgetLimit): Files are in shared module only; Migration35To36 already deleted the table from main app DB
- Issue 2 (limit=50): Code already uses limit = 3 with DB-level profile filtering
- Issue 3 (duplicated filter logic): Already using centralized filterAccountsByProfile() / filterTransactionsByProfile() helpers
- Issue 5 (toggle logic): Replaced with cleaner updateProfileId(newId) with comments
- Issue 7 (schema default): @ColumnInfo(name = "profile_id", defaultValue = "NULL") is correct for a nullable column
- Issue 8 (cache timing): Uses _cachedAccountBalances.filterNotNull() flow pattern
- TransactionItem nit: isEffectivelyBusiness correctly uses profileAccountKeys for inherited account profile

sarim2000:
Awesome, will check this on priority, appreciate this cmazwar and get this merged today

sarim2000:
Tested this locally. The feature works, but I'd hold off on merging for two issues:

### 1. Profile filter is invisible at rest on Home screen (UX bug)

The filter icon is wired into `HomeScreen.kt` inside `actionContent`, which only renders in the collapsed `RegularTopAppBar`. On home screen, the expanded state shows `GreetingCard` (avatar + greeting + its own `MoreHoriz` menu) via `extraInfoCard`, and `LargerTopAppBar` explicitly hides `actionContent` when `isHomeScreen == true` (see `CustomTitleTopAppBar.kt:202-208`). Net effect: until the user scrolls down, the profile filter is not shown anywhere — a brand-new user has no discoverable entry point to the main feature of this PR.

Fix options:
- Add the profile filter icon inside `GreetingCard` (next to or alongside the existing `MoreHoriz` button), or
- Pass the filter through as part of `extraInfoCard` so it renders in the expanded state too.

### 2. Custom profiles can't be created from the UI

The data layer fully supports arbitrary profiles — `ProfileEntity`, `ProfileDao.insert/update/deleteById` (with a guard on ids 1 & 2), `ProfileRepository` — but there's no UI anywhere to create, rename, or delete a profile. Users are locked to the two seeded rows (`PERSONAL_ID = 1`, `BUSINESS_ID = 2`) even though the schema and repository are clearly built for more. That feels half-finished: either add a small profile-management screen, or drop the generic `ProfileEntity` modeling and just use a nullable `isBusiness` column until real multi-profile UX is designed.

### Other notes
- `parser-core/.../CBEBankParser.kt` changes and the deletion of `parser-core/src/test/kotlin/TestCBEBankParser.kt` look unrelated to business/personal profiles — probably worth splitting into a separate PR so this one stays focused.
- DB migrations v37→v40 look clean on a fresh install; didn't test an upgrade path from an existing v36 install yet.

Happy to help patch (1) on top if you'd like — it's a small change in `GreetingCard.kt` + `HomeScreen.kt`.

image

cmazwar:
ill fix the first 1. for the second pardon me, but it's not half-finished, it's intentionally scoped to two system profiles. If custom profiles are needed later, the data layer is ready. No code changes needed. this is already a big feature so it can be foundation for more expansion in future.

PR #302, opened 2026-05-04
feat: user-defined custom SMS parsers with token-tagging UI

### LINKED ISSUE #279: [FEATURE] Custom user defined parsing rule
## 💡 Feature Description
Give an option by which user can parse the messages by using something like regex expressions 

## 🎯 Problem it Solves
It will give user the power to recognise the msgs that are currently going to waste without being recognised 

Kindly refer
(https://github.com/ritesh-kanwar/Cashiro/issues/26)

sarim2000:
## Summary

Closes #279.

Lets users teach PennyWise to read SMS from senders that no built-in bank parser supports — without having to hand-write regex.

**How it works**

- **Settings → Custom Parsers** lists user-defined rules with toggle / edit / delete.
- The **add / edit** screen takes a sample SMS, tokenizes it on whitespace into tappable chips, and pops a sheet when you tap one: *Amount / Merchant / Account # / Expense keyword / Income keyword / Skip*.
- Behind the scenes the editor takes the 1 token before and 1 token after the tagged token as anchors and synthesizes a regex for that field. A live preview shows what the rule would extract from the sample as you tag.
- Built-in `BankParserFactory` parsers still run first — custom rules only fire as a fallback when no built-in parser claims the sender. Existing bank support is unaffected.
- Schema bumped 46 → 47 with a manual migration that creates `custom_parser_rules`. No changes to existing tables.

**Out of scope (deliberately) for this PR**

- Multi-sample-SMS training — one sample per rule.
- Currency auto-detect — user picks (default INR).
- Balance / mandate / subscription extraction.

## Test plan

- [ ] Settings → Custom Parsers opens.
- [ ] Empty state shows "Add custom parser" CTA.
- [ ] FAB opens new-rule editor.
- [ ] Pasting a sample SMS renders chips.
- [ ] Tapping a chip opens the bottom sheet; selecting a tag colors the chip and updates the live preview.
- [ ] Save is disabled until name + sender + display name + amount tag + at least one expense/income keyword are present.
- [ ] After saving, the rule appears in the list.
- [ ] Toggling active off / on persists across screen rotation.
- [ ] An SMS from a sender that **does** match a built-in bank parser still routes to that built-in (custom never shadows real banks).
- [ ] An SMS from a sender that **does not** match any built-in parser, but matches a custom rule's sender pattern, gets parsed using the rule's regex.
- [ ] Edit existing rule loads name / sender / display / currency back into the editor (note: tag chips do not re-hydrate from a saved rule — only the regex does).
- [ ] Delete confirmation dialog removes the rule.

### DIFF
--- a/app/schemas/com.pennywiseai.tracker.data.database.PennyWiseDatabase/47.json
+++ b/app/schemas/com.pennywiseai.tracker.data.database.PennyWiseDatabase/47.json
@@ -1,0 +1,1831 @@
+{
+  "formatVersion": 1,
+  "database": {
+    "version": 47,
+    "identityHash": "d3ae746ea54899f0bf89cd162d07d8d1",
+    "entities": [
+      {
+        "tableName": "transactions",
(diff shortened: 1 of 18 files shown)


sarim2000:
Closing this as stale. It's been ~2 months untouched (+5517), and `parser-core` / the parser architecture have moved substantially since (many new parsers + the currency work landed), so it needs a real rebase and re-review rather than a quick merge — and Greptile currently rates it 2/5.

Not abandoning the idea — the branch is preserved, so this can be reopened/rebased if user-defined custom SMS parsers come back onto the roadmap. Closing for now to keep the open-PR list to things that are actually mergeable.

Issue #308, opened 2026-05-09
Adjust "Spent this month" based on income Refund/Extra Budget


AlphaXyph:
As the title says:

If it's a refund the users would expect the "Spent this month" to get that deducted from the total spent here.

If it's extra budget then it probably don't want any change in "Spent this month"

So only add a condition on "Refund"

PR #313, opened 2026-05-13
fix: apply Refund / Extra Budget on Budgets screen in unified-currency mode

### LINKED ISSUE #312: [BUG] Budgets screen ignores Refund / Extra Budget in unified-currency mode
## 🐛 Bug Description

On the Budgets screen, marking an INCOME transaction with **Budget impact = Refund** (`DEDUCT_SPENT`) does not reduce the linked category's spent amount, and **Budget impact = Extra budget** (`ADD_TO_LIMIT`) does not raise the linked category's displayed budget — **but only when the user is in unified-currency mode** (Settings → Unified currency = ON).

The single-currency Budgets screen path applies both correctly. The unified-mode path silently skips the entire budget-impact logic.

## 📱 Steps to Reproduce

1. Enable **unified currency mode** (Settings → display a single converted currency).
2. Create a Budget group → add a category, e.g. "Shopping", with a limit.
3. Add some EXPENSE transactions in that category for the current month.
4. Add an INCOME transaction for the same month and tag it:
   - **Budget impact: Refund → Shopping**, *or*
   - **Budget impact: Extra budget → Shopping**.
5. Open the Budgets screen.

## ✅ Expected Behavior

- **Refund**: Shopping's "spent" amount drops by the refund amount (floored at zero).
- **Extra budget**: Shopping's displayed budget limit is raised by the income amount.

## ❌ Actual Behavior

- Spent stays the same; budget limit stays the same. The income transaction has no visible effect on the category.

## 🔍 Root Cause

`BudgetGroupsViewModel.convertRawToSummary` (the unified-mode path) is missing the budget-impact reducer that lives in `BudgetGroupRepository.buildSummary` (the single-currency path):

- `app/src/main/java/com/pennywiseai/tracker/presentation/budgetgroups/BudgetGroupsViewModel.kt`
  - Lines 124–135 build `categoryAmounts` only from `transactionType != INCOME`. There is no second pass over INCOME transactions to apply `DEDUCT_SPENT` deductions.
  - Line 186 reads `cat.budgetAmount` directly with no `categoryLimitBoosts` map, so `ADD_TO_LIMIT` never reaches the rendered budget.

Compare with the equivalent loop in `BudgetGroupRepository.buildSummary` (`app/src/main/java/com/pennywiseai/tracker/data/repository/BudgetGroupRepository.kt:347-363` and `:444`) which does both.

## 🔧 Suggested Fix

Mirror the `buildSummary` logic inside `convertRawToSummary`:
1. After populating `categoryAmounts` from non-INCOME txns, add a second pass over INCOME txns with `budgetCategory != null`:
   - `DEDUCT_SPENT` → subtract converted amount from `categoryAmounts[budgetCategory]` (`coerceAtLeast(ZERO)`).
   - `ADD_TO_LIMIT` → accumulate converted amount into a `categoryLimitBoosts` map.
2. When building each `BudgetCategorySpending`, use `convertedBudget + (categoryLimitBoosts[name] ?: ZERO)` instead of `convertedBudget` alone.

Both the deduction and the boost amounts must go through `currencyConversionService.convertAmount(tx.transaction.amount, tx.transaction.currency, displayCurrency)` since this is the unified path.

## 📋 Notes

- Related to issue #308 (which covers the same conceptual gap on the home screen "Spent this month" stat), but this is a separate file and a separate fix.

sarim2000:
## Summary
- The unified-currency Budgets path (`BudgetGroupsViewModel.convertRawToSummary`) was missing the budget-impact reducer that lives in `BudgetGroupRepository.buildSummary` (single-currency path). INCOME transactions tagged with **Refund** (`DEDUCT_SPENT`) had no effect on the linked category's spent total, and **Extra budget** (`ADD_TO_LIMIT`) never raised the displayed budget.
- This commit mirrors the same loop: walk `INCOME` transactions with `budgetCategory != null`, convert each amount to the display currency, then either subtract it from `categoryAmounts` (floored at zero) or accumulate it into a `categoryLimitBoosts` map that is added to the rendered budget.

Fixes #312.

## Test plan
- [x] `./gradlew :app:compileStandardDebugKotlin` — builds clean.
- [ ] Manual on-device check, unified-currency mode ON:
  - [ ] Spend in a category, then mark an INCOME transaction as **Refund → that category**; verify the category's spent amount drops by the refund (floored at 0) and "Spent / Budget" updates accordingly.
  - [ ] Mark an INCOME transaction as **Extra budget → that category**; verify the category's budget limit raises by that amount.
  - [ ] Regression: single-currency mode still works the same.
  - [ ] Regression: INCOME with no budget impact, or impact pointing at a non-existent category, doesn't affect any totals.

## Notes
- This PR is intentionally scoped to the budget-impact reducer. The unified path still includes TRANSFER/INVESTMENT in `categoryAmounts` whereas the single-currency path excludes them — pre-existing inconsistency, not addressed here.
- Issue #308 is a related-but-separate gap on the home screen "Spent this month" stat and will be its own PR.

### DIFF
--- a/app/src/main/java/com/pennywiseai/tracker/presentation/budgetgroups/BudgetGroupsViewModel.kt
+++ b/app/src/main/java/com/pennywiseai/tracker/presentation/budgetgroups/BudgetGroupsViewModel.kt
@@ -5,6 +5,7 @@ import androidx.lifecycle.ViewModel
 import androidx.lifecycle.viewModelScope
 import com.pennywiseai.tracker.data.currency.CurrencyConversionService
 import com.pennywiseai.tracker.data.database.entity.BudgetGroupType
+import com.pennywiseai.tracker.data.database.entity.BudgetImpactType
 import com.pennywiseai.tracker.data.database.entity.TransactionType
 import com.pennywiseai.tracker.data.preferences.UserPreferencesRepository
 import com.pennywiseai.tracker.data.repository.BudgetCategorySpending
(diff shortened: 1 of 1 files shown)


sarim2000:
Closing — going to redo this as a proper refactor that extracts the per-category aggregation into a shared helper in BudgetGroupRepository, so the single-currency and unified-currency paths share one source of truth. The mirror approach in this PR leaves the same kind of duplication that caused #312 in the first place.

PR #389, opened 2026-05-31
feat: prefer transaction description over merchant name on rows (#383)

### LINKED ISSUE #383: [FEATURE] Transaction Display changes
## 💡 Feature Description
* Instead of merchant name the transactions should show description of it has been added.

## 🎯 Problem it Solves
Gives a detailed view of what the transction is for
Because most of the times the merchant name is not known or is some gibberish like it's upi id so it doesn't give any idea of why the transaction is present. 
## 🔧 Proposed Solution
If a description is added description should be shown as a heading of the transaction and not the merchant name.
For the category we can assign a emoji to see it visually for each transaction or we can add the category name on the transaction itself

## 🎨 UI/UX Suggestions (Optional)

Image

sarim2000:
## Why
SMS-parsed merchant strings are often cryptic UPI VPAs (`xyz@okhdfcbank`) that don't help the user remember what a transaction was for. When the user has added a description, that's the most useful label for the row.

## Change
When `transaction.description` is non-blank, use it as the row title; otherwise fall back to the existing merchant-name resolution. To preserve context, the merchant name is surfaced as the first segment of the row's subtitle so the user still sees who the transaction was with.

Applied in both rendering paths:
- **`TransactionItem`** — home recent items, transactions list, search results.
- **`GroupTransactionItem`** — rows in the transaction-group detail screen (previously didn't even show a description; now it leads, with merchant in the secondary line).

## Verification
- `./gradlew :app:compileStandardDebugKotlin` — clean.

## Note on the visual-category suggestion in the issue
`TransactionItem` already renders a category-aware `BrandIcon` on the leading edge (which falls back to a category icon when no brand match exists), so the category is already visually present on the most-used row. The issue's secondary ask was the description-as-heading; that's what this PR delivers. Happy to extend with an explicit category chip on `GroupTransactionItem` as a follow-up if you want it there too.

Closes #383

### DIFF
--- a/app/src/main/java/com/pennywiseai/tracker/presentation/groups/TransactionGroupDetailScreen.kt
+++ b/app/src/main/java/com/pennywiseai/tracker/presentation/groups/TransactionGroupDetailScreen.kt
@@ -456,15 +456,26 @@ private fun AddTransactionToGroupSheet(
                                 horizontalArrangement = Arrangement.SpaceBetween,
                                 verticalAlignment = Alignment.CenterVertically
                             ) {
+                                // Match the description-first heading used on the group's
+                                // own rows so the picker label is consistent with what
+                                // the user will see once the txn is added. (#383)
+                                val txnDescription = txn.description?.takeIf { it.isNotBlank() }
                                 Column(modifier = Modifier.weight(1f)) {
                                     Text(
-                                        txn.merchantName,
+                                        txnDescription ?: txn.merchantName,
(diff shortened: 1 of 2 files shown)


shreyes-shalgar:
@sarim2000 I really want to see the change where the description is brcomes the transaction title as it gives a very clearand comprehensive idea of what the transaction is for instead of just starting at some random upi id. :)

Issue #451, opened 2026-06-08
[FEATURE] Exclude large one-time purchases from spending analytics


rafi-ghanbari:
## 💡 Feature Description
Add the ability to exclude specific transactions from analytics and reports, or mark them as one-time large purchases/assets.

These transactions should remain visible in transaction history while being optionally ignored by spending analytics.

## 🎯 Problem it Solves
Large purchases such as cars, motorcycles, laptops, phones, furniture, or home appliances can significantly distort spending statistics.

For example, if a user buys a motorcycle in a given month, the monthly spending chart, averages, trends, and category analysis become heavily skewed. This makes it difficult to understand actual day-to-day spending behavior.

## 🔧 Proposed Solution
Allow users to mark individual transactions as:

- Exclude from analytics
or
- Large purchase / Asset purchase

Excluded transactions would still be stored normally but would not affect:

- Spending trends
- Monthly averages
- Budget analysis
- Category insights
- AI-generated spending summaries

## 🎨 UI/UX Suggestions (Optional)
Possible approaches:

- A checkbutton in the transaction details screen: "Exclude from analytics"
- A dedicated transaction type such as "Asset Purchase"
- A filter option in reports to include or exclude these transactions

## 📱 Alternatives Considered
Creating a separate category for large purchases is possible, but those transactions would still affect analytics and spending reports. The proposed feature would provide more accurate financial insights.

## 💬 Additional Context
Many personal finance applications allow users to exclude transfers, asset purchases, or exceptional one-time expenses from reports. This helps maintain realistic spending trends while keeping a complete transaction history.

PR #457, opened 2026-06-09
fix(budgets): count investments and auto-categorize them as Investments

sarim2000:
## Problem
Investment transactions were invisible to budgets. An **Investments** budget showed **Spent ₹0** even when matching SIP/investment transactions existed, because the spend math (`BudgetRepository`, `BudgetGroupRepository`, and the budget-groups pace chart) filtered to `EXPENSE` only. The flat "Actual" line on the budget chart was the same bug.

## Changes
- **Budgets count investments.** `EXPENSE` **or** `INVESTMENT` now count toward budget spending and the pace "Actual" line (loans and analytics-excluded transactions still excluded).
- **Investments self-categorize as "Investments"** at every entry point so they land in an investment budget instead of leaking into a spending category via "Others":
  - `SharedCategoryMapping.determineCategory` (parsed/auto)
  - `AddViewModel` (manual add) — also fixes a latent bug where it set the category to the singular **"Investment"**, which matched no budget category
  - `TransactionDetailViewModel` (editing a transaction to Investment type)
- **`MIGRATION_51_52`** backfills existing `INVESTMENT` transactions stuck in `Others`/`Investment` → `Investments`. Deliberately-chosen categories (e.g. `Banking`) are left untouched.

## Verification (emulator)
- Investment in "Others" → **Spending** budget went ₹500 → **₹6,500** (now counts).
- Re-categorized to "Investments" → **Investments** budget ₹0 → **₹6,000 / ₹15,000 (40%)**, Spending fell back to ₹500.
- Edit flow: selecting **Investment** type flips the category to **Investments** live.

## Notes
- Budget **groups** still don't honor `excludedFromAnalytics` (#451 only added that filter to the legacy `BudgetRepository`) — out of scope here, flagging for follow-up.

### DIFF
--- a/app/schemas/com.pennywiseai.tracker.data.database.PennyWiseDatabase/52.json
+++ b/app/schemas/com.pennywiseai.tracker.data.database.PennyWiseDatabase/52.json
@@ -1,0 +1,1747 @@
+{
+  "formatVersion": 1,
+  "database": {
+    "version": 52,
+    "identityHash": "a055fe280d3ddb19f45cda61ee0f718d",
+    "entities": [
+      {
+        "tableName": "transactions",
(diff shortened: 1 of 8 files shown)


sarim2000:
Superseded by #467 (type-aware budgets) — fixes the root design flaw instead of working around it with forced categorisation. Closing in favour of the redesign.

PR #487, opened 2026-06-15
fix(accounts): derive manual/cash balance from transactions (#469, #470)

### LINKED ISSUE #469: Cash account not updating balance correctly
## 🐛 Bug Description

Cash account (not tracked by SMS) don't update correctly when you add expenses or income to the account in a date different than "today". The balance remain unchanged. It works fine if you add expenses or income for "today". 

## 📱 Steps to Reproduce
1. Go to 'Main'
2. Click on '+'
3. Add an expense or income to cash account set transaction date to a date different to 'todays' date
4. See that cash balance will not change 

## ✅ Expected Behavior

Balance should change to reflect income or expenses incurred in the past.

## 📋 Device Information
- **Device**: Google Pixel 7 Pro
- **Android Version**: Android 15
- **App Version**: 2.15.60

### LINKED ISSUE #470: Cash balance don't update after deleting an entry
## 🐛 Bug Description

When deleting a entry to a cash account (not tracked by SMS) the balance remain unchanged. 

## 📱 Steps to Reproduce
1. Go to 'Main'
2. Click on '+'
3. Add an expense or income to cash account
4. See balance change to cash account
5. Delete the entry
7. See error where the balance will remain unchanged even after deleting the entry

## 📸 Screenshots

Balance should change after an entry is deleted. The app should calculate the balance according to the records entered by the user. 

## 📋 Device Information
- **Device**: Google Pixel 7 Pro
- **Android Version**: Android 15
- **App Version**: 2.15.60

sarim2000:
Fixes the two cash-account balance bugs by changing how manual/cash balances are computed.

## Root cause
A manual/cash account's balance was maintained as an incremental `latest ± amount` snapshot stamped at **the transaction's date**. So anything retroactive broke it:
- **#469** — a back-dated add wrote a row in the past → never became "latest" → balance unchanged.
- **#470** — delete never reversed the delta.
- (latent) amount/date edits drifted.

## Fix — derive instead of nudge
A manual account's balance = **opening + Σ(its transactions)**, recomputed on every add / edit / delete into a single `MANUAL` row refreshed to `now`.

- New repo API: `recomputeManualBalance`, `ensureManualOpening`, `setManualCurrentBalance`, `seedManualAccount` (injects `TransactionDao` for the signed sum — INCOME +, else −, matching the prior convention).
- A stable **OPENING** anchor row (early timestamp, so it never wins "latest") stores the opening. **No schema change / migration** — `OPENING` is just a new `source_type` string value.
- **Upgrade-safe:** for accounts created before this, the opening is derived lazily from a *pre-mutation* snapshot (`opening = currentBalance − Σtxns`), so the displayed balance never jumps on upgrade and the triggering change is still reflected. (The pre-mutation timing is why `ensureManualOpening` is called *before* the insert/delete/update in each write path.)
- **"Update balance"** back-solves the opening (option b): you still type your *current* balance; future transactions adjust from there.
- **SMS accounts and credit cards are excluded** (balance comes from the bank / has inverted semantics) — their paths are untouched.

## Verification (emulator)
1. Created a cash account "Wallet" $1,000.
2. Added a **back-dated** (Jun 5) $200 expense assigned to Wallet → **$800** ✓ (#469 — old behavior left it at $1,000).
3. Deleted that expense → **$1,000** ✓ (#470).

## Follow-up (not in this PR)
- The balance-history **chart** for manual accounts isn't yet derived from opening + running Σ; it still plots stored rows. Cosmetic, can be a separate change.

Closes #469
Closes #470

### DIFF
--- a/app/src/main/java/com/pennywiseai/tracker/data/database/dao/AccountBalanceDao.kt
+++ b/app/src/main/java/com/pennywiseai/tracker/data/database/dao/AccountBalanceDao.kt
@@ -223,4 +223,43 @@ interface AccountBalanceDao {
 
     @Query("UPDATE account_balances SET alias = :alias WHERE bank_name = :bankName AND account_last4 = :accountLast4")
     suspend fun setAccountAlias(bankName: String, accountLast4: String, alias: String?): Int
+
+    // ─────────────────────────────────────────────────────────────────────────
+    // Manual/cash account balance recompute support.
+    //
+    // A manual account's displayed balance = opening + Σ(its transactions). The
+    // opening is stored in a single OPENING row at an early timestamp (so it never
+    // wins "latest by timestamp"); the displayed value lives in a single MANUAL row
+    // that recompute refreshes. See AccountBalanceRepository.recomputeManualBalance.
+    // ─────────────────────────────────────────────────────────────────────────
+
+    @Query("""
+        SELECT * FROM account_balances
+        WHERE bank_name = :bankName AND account_last4 = :accountLast4
+        AND source_type = 'OPENING'
+        ORDER BY timestamp ASC
+        LIMIT 1
+    """)
+    suspend fun getOpeningRow(bankName: String, accountLast4: String): AccountBalanceEntity?
+
+    @Query("""
+        SELECT * FROM account_balances
+        WHERE bank_name = :bankName AND account_last4 = :accountLast4
+        AND source_type = 'MANUAL'
+        ORDER BY timestamp DESC
+        LIMIT 1
+    """)
+    suspend fun getManualCurrentRow(bankName: String, accountLast4: String): AccountBalanceEntity?
+
+    @Query("""
+        SELECT MIN(timestamp) FROM account_balances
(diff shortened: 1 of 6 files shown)


sarim2000:
@greptileai please re-review — the two findings in the latest summary were already addressed in commit `b21ae43f` and are present in the reviewed HEAD (`142082cf`):

- **`withTransaction` on `recomputeManualBalance`** — wrapped at `AccountBalanceRepository.kt:327` (along with `ensureManualOpening`/`setManualCurrentBalance`/`seedManualAccount`/`updateManualBalanceAndCurrency`).
- **Spurious `TRANSACTION` row on account-change edits** — guarded out for manual accounts at `TransactionDetailViewModel.kt:668` (`shouldRunSingleAccount && !isManualAccount(...)`).

The only intentionally-non-transactional path is the add flow's `ensureManualOpening → insert txn → recomputeManualBalance`, which spans layers; each step is atomic and idempotent so a crash between them self-heals on the next mutation.

PR #515, opened 2026-06-21
fix(rules): allow Smart Rules to set Transaction Type as Expense

saadaltabari:
## Summary

Smart Rules could not reliably set or match **Transaction Type → Expense**. The rule editor showed “Outgoing” instead of “Expense”, kept incompatible operators when switching from Amount to Transaction Type (for example `LESS_THAN`), and duplicated fragile TYPE pre-filter logic across the repository and engine.

This change centralizes transaction-type rule helpers, uses proper `is` / `is not` operators for TYPE conditions, resets operators when the condition field changes, validates TYPE values, and adds unit tests for EXPENSE matching and SET TYPE actions.

## Type of change

- [x] fix: Bug fix
- [x] test: Add/update tests

## How to test

1. Open **Settings → Smart Rules → Create Rule**.
2. Under **Then**, choose **Set Field → Set Transaction Type** and select **Expense**; save the rule.
3. Add a condition **Transaction Type → is → Expense** (optionally combined with another condition); confirm Save is enabled and the preview shows “Expense”.
4. Enable the rule and verify it applies on matching transactions.
5. Run:
   ```bash
   ./gradlew :app:testStandardDebugUnitTest --tests "com.pennywiseai.tracker.domain.service.RuleEngineTest" --tests "com.pennywiseai.tracker.domain.model.rule.RuleConditionTest"
   ```

## Breaking changes

No breaking changes.

## Checklist

- [x] Tests added/updated
- [x] Targeted unit tests pass locally

Made with [Cursor](https://cursor.com)

### DIFF
--- a/app/src/main/java/com/pennywiseai/tracker/data/repository/RuleRepositoryImpl.kt
+++ b/app/src/main/java/com/pennywiseai/tracker/data/repository/RuleRepositoryImpl.kt
@@ -39,62 +39,17 @@ class RuleRepositoryImpl @Inject constructor(
         val allActiveRules = ruleDao.getActiveRules()
         return allActiveRules
             .map { entity -> entityToRule(entity) }
-            .filter { rule ->
-                // Check if rule has no TYPE condition (applies to all) or matches the type
-                val hasTypeCondition = rule.conditions.any { it.field == TransactionField.TYPE }
-                if (!hasTypeCondition) {
-                    true // Rule applies to all transaction types
-                } else {
-                    // Check if any TYPE condition matches
-                    rule.conditions.any { condition ->
-                        condition.field == TransactionField.TYPE &&
-                        when (condition.operator) {
-                            ConditionOperator.EQUALS -> condition.value.equals(type.name, ignoreCase = true)
-                            ConditionOperator.IN -> condition.value.split(",")
-                                .map { it.trim() }
-                                .any { it.equals(type.name, ignoreCase = true) }
-                            ConditionOperator.NOT_EQUALS -> !condition.value.equals(type.name, ignoreCase = true)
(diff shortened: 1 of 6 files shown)


sarim2000:
Nice fix, and the `getActiveRulesByTypes` ANY-match correction is right. One active defect to fix first: in `CreateRuleScreen`, the transaction-type **value chip's `onClick` always forces `operator = EQUALS`**, so a "Transaction Type **is not** Expense" condition can't be saved through the UI (even though the engine/repo now handle NOT_EQUALS correctly). Please preserve the selected operator there.

sarim2000:
Thanks @saadaltabari — the EXPENSE-type fix and helper centralisation look good (Greptile 4/5). One real thing it caught: in `ConditionFieldSelector` (CreateRuleScreen, ~line 874) the type-value chip `onClick` always sets `operator = EQUALS`, so an "is not" condition silently saves as "is". Looks like a one-line removal.

It's fallen behind `main` and now conflicts — a rebase when you have a moment would clear that and re-trigger Greptile.

Gentle housekeeping heads-up: if there's no update in ~a week we might close it just to keep things tidy — no rejection implied, happy to reopen the moment you're back on it. 🙏

sarim2000:
Thanks! One thing from the automated review before merge:

- **[P2]** `CreateRuleScreen.kt:787` — the `LaunchedEffect` operator-reset scope is limited to the TYPE field, so other fields can be left with a stale operator.

Could you widen the reset scope so it stays consistent across fields?

sarim2000:
Closing this as stale — it has been inactive for several weeks with the review feedback above unaddressed. No hard feelings, and thanks for the contribution! Please feel free to reopen (or open a fresh PR) once it is rebased on the latest `main` and the flagged items are addressed, and I will take another look.

PR #554, opened 2026-06-26
fix(home): stop the This-month summary from crushing long amounts

sarim2000:
## What

From the same MZN report: the home "This month" summary rendered the **Saved** amount vertically, one character per line (`M Z N 5 2 , 2 4 5 . 2 6`).

## Root cause

The summary is a single `Row` with `Arrangement.SpaceEvenly` holding up to **four** `SummaryItem`s — Income, Expenses, **Lent**, Saved (Lent only shows when you've lent money this month). With four columns and long currency strings (the `MZN` 3-char prefix on big numbers), the row runs out of width and crushes the last column to ~1 char wide. `SummaryItem`'s value `Text` had no `maxLines`/`softWrap`, so it broke at every character instead of staying on one line.

3 columns fit fine; the 4th column is what overflowed — which is why it shows for users with a Lent total.

## Fix

- Summary row → `FlowRow`, so four items **wrap to a second line (2×2)** instead of crushing the last column.
- `SummaryItem` value → `maxLines = 1`, `softWrap = false` (an amount never breaks mid-number).
- Per-account row → `account.formatBalance()` (the account's own, pre-converted currency) so an un-convertible balance reads `$400`, not a raw amount mislabelled with the display currency.

## Verification

Emulator, MZN display, all four columns present → clean 2×2 grid, every amount on one line. (See before/after.)

## Note

Independent of the conversion fix in #553; both came from the same MZN/USD report.

### DIFF
--- a/app/src/main/java/com/pennywiseai/tracker/ui/components/cards/BalanceCard.kt
+++ b/app/src/main/java/com/pennywiseai/tracker/ui/components/cards/BalanceCard.kt
@@ -10,6 +10,8 @@ import androidx.compose.foundation.isSystemInDarkTheme
 import androidx.compose.foundation.layout.Arrangement
 import androidx.compose.foundation.layout.Box
 import androidx.compose.foundation.layout.Column
+import androidx.compose.foundation.layout.ExperimentalLayoutApi
+import androidx.compose.foundation.layout.FlowRow
 import androidx.compose.foundation.layout.Row
 import androidx.compose.foundation.shape.CircleShape
 import androidx.compose.foundation.layout.Spacer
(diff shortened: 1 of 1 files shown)


sarim2000:
Merging at 4/5: the remaining Greptile comment ('converted balance shows native currency label') is a verified false positive.

`uiState.accountBalances` is produced solely by `convertAccountEntities`, which sets `balance` and `currency` in the same `.copy()` — so an entity is either (converted balance + target currency) or (native balance + native currency). A converted amount with a native label is structurally impossible, so `account.formatBalance()` always pairs the amount with its true currency. The code this replaced (`formatCurrency(balance, selectedCurrency)`) was the actual mislabel source. Layout fix verified on the emulator (MZN, 4 columns → clean 2x2 grid).

PR #572, opened 2026-07-04
fix(analytics,widget): net refunds out of spend totals

sarim2000:
## Problem
A refund (an `INCOME` transaction with `budgetImpactType = DEDUCT_SPENT`) reverses a prior expense. The **Transactions** page, **Home** "Spent this month" card and **Budgets** already subtract it — but two surfaces didn't, so a refund left them showing **gross** spend:

- **Analytics** — total, top categories, average, trend, and the merchant/account breakdowns all showed pre-refund amounts.
- **Recent Transactions widget** — "Total spend this month" showed pre-refund amount.

Reported with: ₹20 + ₹24 spent, ₹12 refunded on the food category → everywhere *should* read ₹32, but Analytics and the widget showed ₹44.

## Root cause
Both surfaces compute their own spend sum and never accounted for refunds. Analytics is worse because it queries **EXPENSE-only** rows, so the refund (an INCOME txn) never even entered the pipeline.

## Fix
- **Analytics** (`AnalyticsViewModel`): in the Expense view, pull refunds back in from the already-loaded (untyped) list and net them out of **every** spend surface, floored at zero — headline total, category (by `budgetCategory`), account (by account key), merchant (when the refund carries the same merchant name), and the spending trend (subtracted on the day it occurred). Average derives from the total. Income/other type views are untouched.
- **Widget** (`RecentTransactionsWidgetUpdateWorker`): subtract refunds from "Total spend this month", matching the Home card.

## Verification (emulator, seeded 20 + 24 expense, 12 refund on food)
| Surface | Before | After |
|--------|--------|-------|
| Analytics total | ₹44 | **₹32** ✅ |
| Food & Dining category | ₹24 | **₹12** ✅ |
| Education category | ₹20 | ₹20 ✅ |
| Average / day | ₹22 | **₹16** ✅ |
| Trend peak | ₹44 | **₹32** ✅ |
| By Account % | **137%** | **100%** ✅ |
| Cafe merchant (refund shares merchant) | ₹24 | **₹12** ✅ |
| Widget "Total spend this month" | ₹44 | **₹32** ✅ (read from datastore) |

App unit suite green; `:app:compileStandardDebugKotlin` clean.

### DIFF
--- a/app/src/main/java/com/pennywiseai/tracker/ui/screens/analytics/AnalyticsViewModel.kt
+++ b/app/src/main/java/com/pennywiseai/tracker/ui/screens/analytics/AnalyticsViewModel.kt
@@ -304,6 +304,45 @@ class AnalyticsViewModel @Inject constructor(
                     }
                 }
 
+                // Net "Refund" income (INCOME + DEDUCT_SPENT) out of Expenses so EVERY
+                // spend surface — total, category, account, merchant and the trend —
+                // agrees with the Home card and the Transactions page (both already net
+                // of refunds). A categorised refund subtracts from its budgetCategory;
+                // it also shrinks the headline total, its account, its merchant (when
+                // the refund carries the same merchant) and the trend on its day. All
+                // spend surfaces are floored at zero. Refunds live in
+                // allTransactionsWithSplits (loaded untyped) but were dropped by the
+                // EXPENSE type filter above, so pull them back in here. Only the Expense
+                // view is adjusted; Income/other views are untouched.
+                val refundByAccount = mutableMapOf<String, BigDecimal>()
+                val refundByMerchant = mutableMapOf<String, BigDecimal>()
+                val refundTransactions =
+                    mutableListOf<com.pennywiseai.tracker.data.database.entity.TransactionEntity>()
+                if (filterState.typeFilter == TransactionTypeFilter.EXPENSE) {
+                    for (tw in allTransactionsWithSplits) {
+                        val tx = tw.transaction
+                        if (tx.loanId != null || tx.excludedFromAnalytics) continue
+                        if (tx.transactionType != com.pennywiseai.tracker.data.database.entity.TransactionType.INCOME ||
+                            tx.budgetImpactType != com.pennywiseai.tracker.data.database.entity.BudgetImpactType.DEDUCT_SPENT
+                        ) continue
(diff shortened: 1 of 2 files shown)


sarim2000:
Addressed the review:

**#1 (total could drop below the breakdown rows) — fixed.** Each refund is now capped at its category's spend this period, and that single *effective* amount is netted from the total, its category, account, merchant and the trend (via a per-day map). So an orphaned refund (category with no expense here) or the excess of an over-refund contributes nothing, and `total == Σ category == Σ account` always holds. Verified on emulator: 20 Education + 10 Food + 12 refund on Food → total **20** = Education 20 + Food 0, By Account **100%**, trend peak **20**, Cafe **0**.

**#2 (widget subtracts refunds Analytics excludes) — intentional, not a bug.** The widget mirrors the **Home 'Spent this month' card** (`computeBreakdownByCurrency`), not Analytics. Home includes `excludedFromAnalytics` rows in both its gross expenses *and* its refund subtraction (that opt-out is Analytics-only, per #451), and the widget's gross does the same — so the widget is internally consistent with the surface it's meant to match. Gating its refund on `excludedFromAnalytics` would make it diverge from the Home card it mirrors.

sarim2000:
Merging at Greptile 4/5 intentionally. The two remaining points are resolved-by-design, not open defects:

1. **"Breakdown percentages don't reconcile with the net total."** This is the deliberate trade-off. Analytics now nets the *full* refund from the spend total so it stays consistent with the Home card, widget and Transactions page (all subtract-always). On an over-refund the floored category rows can sum above the total — which is mathematically unavoidable unless we understate the refund (the old cap did that, and made Analytics show ₹20 while Home showed ₹18). The only visible symptom (>100% share) is clamped. "total == Σ rows" and "consistent across screens" are mutually exclusive here; we chose consistency.

2. **"Widget subtracts refunds excluded from analytics."** Verified false positive. The widget's "Spent this month" mirrors the Home card (`computeBreakdownByCurrency`), which intentionally does not filter `excludedFromAnalytics` — it counts excluded *expenses* too. Subtracting an excluded refund is the correct counterpart; filtering only the refund would diverge from Home.

Verified on the emulator across five scenarios (normal, over-refund, orphaned, unified-currency, multi-month trend). Same call as #554.

PR #579, opened 2026-07-06
feat(currency): add LKR (Sri Lankan Rupee) display support

Chevindu:
## Summary

Sri Lankan bank parsers (Sampath, Nations Trust, National Savings) already emit "LKR", but the app-level display layer never mapped it, so amounts showed a literal "LKR" prefix and LKR was absent from the currency pickers.

- CurrencyFormatter: map LKR to the "Rs" symbol (matching the existing Sri Lanka CountryMeta in SupportedBanksDocTest) and an en-LK locale.
- CurrencyUtils: list LKR in getAllSupportedCurrencies() so it appears in the Settings base/display currency dropdowns.

LKR is a standard 2-decimal, western-grouped currency, so no changes to the three-decimal or Indian-notation sets.

## Type of change

- [x] feat: New feature

## Screenshots/Recordings (optional)

N/A - no new UI; LKR now appears as a row in the existing currency pickers.

## How to test

1. Open Settings → Default Currency dropdown → confirm LKR (Rs) is listed and selectable.
2. Add an account (or receive an SMS from a Sri Lankan bank) with currency LKR → confirm amounts render as `Rs 1,234.50`, not `LKR1,234.50`.
3. Change the display currency to LKR → confirm totals across the app show the Rs symbol with `1,234,567.89` western grouping.

## Breaking changes

- [x] No breaking changes

## Related issues

Attempts to answer https://github.com/sarim2000/pennywiseai-tracker/discussions/578.

## Checklist

- [x]  Tests added/updated

### DIFF
--- a/app/src/main/java/com/pennywiseai/tracker/utils/CurrencyFormatter.kt
+++ b/app/src/main/java/com/pennywiseai/tracker/utils/CurrencyFormatter.kt
@@ -134,11 +134,22 @@ object CurrencyFormatter {
     )
 
     /**
+     * Currencies whose ISO code is valid (so `Currency.getInstance()` succeeds and
+     * the try/catch fallback below never triggers), but whose *locale-derived*
+     * symbol collides with another currency's — e.g. `en-LK` can format LKR with a
+     * plain "Rs", identical to PKR's symbol, making mixed rupee amounts (as in
+     * [formatByCurrency]) indistinguishable. These bypass `NumberFormat`'s
+     * currency-instance symbol entirely and always render the explicit
+     * [CURRENCY_SYMBOLS] value instead.
+     */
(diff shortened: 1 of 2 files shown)


sarim2000:
Thanks for this, @Chevindu — and nice catch on the PKR/LKR symbol collision. 🙏

Heads up: #581 landed just ahead of this and added the broader currency list (which is why this now shows conflicts), but it shipped LKR as a plain "Rs" — identical to PKR. Your `රු` symbol + the collision guard/test are the correct fix and I'd like to take them.

Could you rebase onto `main`? The conflict is only in the same few lines #581 touched, so it should be a light one. Happy to help if anything's unclear.

Chevindu:
Thanks for the heads up @sarim2000, I rebased my branch to `main` and updated the PR. Please have a look if it looks okay.

Issue #570, opened 2026-07-02
[BUG]Account allocation and balance updation for Subscription transaction


shreyes-shalgar:
## 🐛 Bug Description
1. For subscription transaction their is no way yo allocate an account to a particular subscription
2. Even if I allocate an account manually the amount is not debited/credited from the account balance

## 📱 Steps to Reproduce
1. Go to subscriptions
2. Click on mark as paid
4. See error 1.
5. Edit the transaction created and add an account
6. Check balance of that account
7. See error 2.

## ✅ Expected Behavior
Subscriptions should have a way to assign a default account
And the balanceShould be updated accordingly

## 📸 Screenshots
 
<img width="720" height="1600" alt="Image" src="https://github.com/user-attachments/assets/6ff2f882-1cb0-4eed-af4c-b603e3e71198" />

PR #580, opened 2026-07-07
fix(txn): edit picker sets bankName so account balance updates (#566, #570)

### LINKED ISSUE #566: [BUG] Balance of individual banks not updating after transactions are added. it stays the same we set previously while adding bank.
## 🐛 Bug Description
Individual bank accounts balance amount is not updating manually or automatically, it stays the same no matter how many transactions we do, and self transfer option doesnt work when i select transfer in the add expense section it doesnt show who we are paying either me or someone else, i have to go save it and edit it again to get that option. 

## 📱 Steps to Reproduce
2. Click on '...'
3. See error

## ✅ Expected Behavior
Balance should automatically update after every transactions so that i can keep track of all my money and cashflow. 

## 📸 Screenshots
Cant upload anything as i have already typed so much and dont want to go back to take screenshots forgive me, maybe ill mail you later. 

## 📋 Device Information
- One plus nord CE 3 5G, everything is up to date even the app from fdroid repo

sarim2000:
## Problem
Editing a transaction to assign an account **did not update that account's balance** — reported independently in **#566** and **#570** (a second reporter confirmed #566's exact repro).

## Root cause
The account picker in `TransactionDetailScreen.AccountNumberField` set only `accountNumber` (the last-4) on select, never `bankName`:

```kotlin
onClick = {
    selectedAccount = account.displayName
    onAccountNumberChange(account.accountLast4)   // ← only the last-4
}
```

Accounts are keyed by **(bankName, accountLast4)**. So a transaction that started on a *different* bank kept its stale `bankName`:
- a **subscription** mark-as-paid row (`bankName` = the subscription's), or
- a **"Manual Entry"** row (manual add without an account).

`saveChanges()` then called `getLatestBalance(staleBank, newLast4)`, missed, and silently skipped the balance update. The balance never moved.

## Fix
Propagate the selected account's `bankName` alongside its last-4:
- `TransactionDetailViewModel.updateBankName()` sets the transaction's bank.
- `AccountNumberField` gains an optional `onBankNameChange`, fired from the dropdown `onClick` next to `onAccountNumberChange`. **Only the single-account field wires it**; the transfer From/To fields (which key on `fromAccount`/`toAccount`) pass `null` and are unchanged.

The manual **add** flow was already correct — it passes a full `AccountBalanceEntity` (both `bankName` + `accountLast4`) to `AddTransactionUseCase`.

## Verification (emulator, ground-truth via DB)
1. Added a ₹100 expense with **no account** → `bank_name = "Manual Entry"`, `account_number = NULL`. Kotak ••7777 baseline = **₹2000**.
2. Edited it → assigned **Kotak ••7777** → Save.
3. Result: `bank_name = "Kotak Bank"`, `account_number = 7777`, and a new balance row **₹1900** (2000 − 100). ✅

Before the fix, `bank_name` stayed `"Manual Entry"`, the lookup missed, and the balance stayed ₹2000.

## Scope
- Resolves **#566** and the balance half of **#570**.
- Assigning a *default* account to a subscription so mark-as-paid auto-hits it (**#570 part 1**) remains a separate enhancement.

Closes #566

### DIFF
--- a/app/src/main/java/com/pennywiseai/tracker/presentation/transactions/TransactionDetailScreen.kt
+++ b/app/src/main/java/com/pennywiseai/tracker/presentation/transactions/TransactionDetailScreen.kt
@@ -1304,6 +1304,7 @@ private fun EditableExtractedInfoCard(
                 AccountNumberField(
                     accountNumber = transaction.accountNumber,
                     onAccountNumberChange = { viewModel.updateAccountNumber(it) },
+                    onBankNameChange = { viewModel.updateBankName(it) },
                     viewModel = viewModel
                 )
             }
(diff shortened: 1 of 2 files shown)


jagadeesh-jj:
Great thanks....

PR #591, opened 2026-07-09
feat: add Pasargad/Wepod bank parser and tests

rafi-ghanbari:
## Summary
Adds support for parsing SMS transaction notifications from Pasargad / Wepod Bank (Iran).

## Type of change

- [ ] feat: New feature
- [ ] fix: Bug fix
- [x] docs: Documentation update
- [x] test: Add/update tests

## How to test

 Run the Pasargad Bank parser unit tests to verify parsing rules:
   ```bash
   ./gradlew :parser-core:jvmTest --tests "com.pennywiseai.parser.core.bank.PasargadBankParserTest"


sarim2000:
Thanks for this, @rafi-ghanbari — parser and tests look solid on the happy path. 🙏 Greptile flagged three small hardening points I'd like tightened before merge (all minor, none blocking the core logic):

1. **`extractAccountLast4` fallback** (`PasargadBankParser.kt:62`) — when `pureNumeric.length < 4` it returns `fullAccount` (a dot-separated string like `1.2.3`) instead of `null`. The contract is "last 4 digits or null," so return `null` in that case (Greptile posted a suggestion inline).

2. **OTP/promo guard bypass** (`PasargadBankParser.kt:29`) — the `isTransactionMessage` override skips `BaseIranianBankParser`'s OTP/promo rejection (`otp`, `رمز یکبار مصرف`, `کد تایید`). Please delegate to / call `super` so those guards still apply — cheap insurance against a future false positive.

3. **OTP ignore test** (`PasargadBankParserTest.kt:104`) — the two ignore-case messages don't contain `مانده:` or a signed amount, so they'd never match regardless; the test doesn't actually exercise the OTP-rejection path. Once #2 is in, add an ignore case that *would* otherwise look like a transaction but contains an OTP keyword.

Once those are addressed it's good to go. Appreciate the contribution!

sarim2000:
Thanks for the parser + tests! The automated review flagged a few things to tighten up before merge:

- **[P2]** `PasargadBankParser.kt:62` — inconsistent fallback in `extractAccountLast4`.
- **[P2]** `PasargadBankParser.kt:29` — the OTP/promo filter isn’t inherited (promotional/OTP SMS may be parsed as transactions).
- **[P2]** `PasargadBankParserTest.kt:104` — the ignore-test messages are trivially non-matching, so they don’t really prove the filter works.

Could you address these, especially the OTP/promo filtering + a real negative test?

sarim2000:
Thanks for adding Pasargad/Wepod — nice to get Iran coverage! CI is green and the parser is nicely isolated. Two small things from review before merge:

1. **`extractAccountLast4` fallback** (~line 62): when `pureNumeric.length < 4` it returns `fullAccount` (a dot-separated string like `"1.2.3"`) instead of `null`. The contract is "last 4 digits or null", so please return `null` in that branch.
2. **OTP/promo guard** (~line 29): the `isTransactionMessage` override bypasses `BaseIranianBankParser`'s OTP/promo rejection (`otp`, `رمز یکبار مصرف`, `کد تایید`). Please delegate to / preserve those guards so an OTP that happens to carry a number can't be parsed as a txn. A test with a realistic Pasargad-shaped OTP (one that contains `مانده:` or a signed amount) would lock it in — the current ignore-test strings don't actually exercise the rejection path.

Both are quick; once addressed I'll re-run and merge. 🙂

PR #609, opened 2026-07-15
fix(parser): IndusInd credit-card refund merchant + card detection (#486)

### LINKED ISSUE #486: IndusInd credit card — available-limit tracking & refund handling.
🏦 Bank Information
Bank Name: IndusInd Bank (Credit Card)
Country: India
📱 SMS Examples
�

Example 1 — Credit-card spend
Sender: 
INR ***.** spent on IndusInd Card XX**** on 14-06-2026 04:21:45 pm at INSTAMART. Avl Lmt: INR ******.**. To dispute, call 18602677777/SMS BLOCK **** to 5676757

Example 2 — Credit-card refund / return
Sender: 
Dear Customer, refund of INR ** from Swiggy Limited Banga has been credited to your IndusInd Bank Credit Card XX**** on 11-Jun-26 and the refund amount has been adjusted against the outstanding on your card account - IndusInd Bank

📋 Transaction Types Needed
[x] Card transactions (credit card)
[x] Available limit tracking
[x] Refund / return handling
🐛 Summary
For an IndusInd credit card (latest main, v2.15.x), credit-card spends are not recognised as CREDIT, available-limit tracking never runs even though the SMS contains Avl Lmt, and refunds/returns are not netted against credit-card spend. Root cause and surgical fixes below — happy to open a PR with test cases.
🔎 Issue 1 — IndusInd credit-card spends are typed EXPENSE, so available-limit tracking never runs
IndusIndBankParser.extractTransactionType() maps "spent" → TransactionType.EXPENSE, so a credit-card purchase is never typed CREDIT. In BankParser.parse(), extractAvailableLimit() is gated behind type == CREDIT, so the Avl Lmt value present in every spend SMS is never read. The common user workaround (a Smart Rule SET TYPE = CREDIT) runs after parsing and only relabels the type — it can't recover the limit, which was never extracted.
Suggested fix (IndusIndBankParser): detect credit-card context and return CREDIT. Card spends say spent on IndusInd Card ... Avl Lmt; account debits say debited from A/C ... Avl Bal. A reliable signal, evaluated before the generic "spent" branch:
Kotlin
lower.contains("card") && lower.contains("avl lmt") -> TransactionType.CREDIT

🔎 Issue 2 — The available-limit regex doesn't match IndusInd's INR format
Even once the type is CREDIT, BankParser.extractAvailableLimit() would still miss this SMS. The patterns require Rs:
Kotlin
Regex("""Avl\s+Lmt:?\s*Rs\.?\s*([0-9,]+(?:\.\d{2})?)""", RegexOption.IGNORE_CASE)
IndusInd prints Avl Lmt: INR ******.** — no Rs, so no match.
Suggested fix: broaden the currency token across the limit patterns from Rs\.? to (?:Rs\.?|INR|₹), or add an IndusInd-specific extractAvailableLimit override.
🔎 Issue 3 — Credit-card refunds aren't netted against credit-card spend
The refund SMS parses as INCOME (via "credited"), which is correct, and for a credit-card account SmsTransactionProcessor already reduces the outstanding balance on INCOME (isCreditCard && INCOME -> balance - amount), so the available limit is restored properly. Two gaps remain:
The monthly credit-card figure (HomeViewModel.currentMonthCreditCard) is a blind sum of CREDIT rows with no refund offset, so a return never reduces "credit card spent this month".
The existing refund-netting (budgetImpactType = DEDUCT_SPENT in computeBreakdownByCurrency) only subtracts from the EXPENSE bucket, never the CREDIT bucket. And DEDUCT_SPENT isn't a settable Smart Rule field, so users can't apply it to auto-parsed transactions anyway.
Suggested direction: mirror the EXPENSE refund-netting for the CREDIT bucket (subtract credit-card refunds from currentMonthCreditCard), and consider auto-tagging credit-card refunds so they reduce credit spend rather than appearing as income. Note that fixing Issue 1 (native CREDIT typing) also removes the need for the user Smart Rule — which is what currently causes returns to be mislabelled as credit-card spend.
🧭 Design note (to avoid a double-count)
IndusInd's SMS reports the available (remaining) limit, but the card model derives available = creditLimit - balance, treating the stored value as the total limit. Feeding Avl Lmt straight into the limit field would subtract the outstanding from an already-net number. Two clean options: (a) treat the SMS value as the available figure and surface it directly, or (b) back-calculate total = avlLmt + outstanding. Worth deciding before wiring it in.
🧹 Minor (general robustness)
In BankParser.extractTransactionType(), "refund"/"cashback" are checked after the expense keywords ("debited", "spent", "purchase", …). Any refund SMS that also contains a spend word would mis-type as EXPENSE for banks relying on the base. Hoisting the refund/reversal check above the expense branch (and adding "reversal"/"reversed") would make refund detection more robust across banks.
📋 Environment
App Version: 2.15.x (latest main)
Module: parser-core (IndusIndBankParser, BankParser), app (HomeViewModel, SmsTransactionProcessor)

sarim2000:
Closes #486 (parser-core scope).

## Problem
For an IndusInd **credit card**, the reporter saw: spends not recognised as CREDIT, `Avl Lmt` tracking never running, and refunds mishandled.

On current `main`, spends already type as `CREDIT` and `Avl Lmt: INR ...` is already captured into `creditLimit` (base `extractAvailableLimit` accepts the `INR` token). The remaining gap was the **refund path**:

- refund SMS produced a garbage merchant (`your IndusInd Bank Credit Card XX1234`), and
- was not flagged as a card txn (`isFromCard=false`) — the trailing `on your card account` phrase makes the base `detectIsCard` bail.

Sample refund SMS:
> Dear Customer, refund of INR 250 from Swiggy Limited Banga has been credited to your IndusInd Bank Credit Card XX1234 on 11-Jun-26 and the refund amount has been adjusted against the outstanding on your card account - IndusInd Bank

## Fix
- **`extractMerchant`** — parse the `refund of INR <amt> from <Merchant> has been credited` shape and strip the trailing legal-entity/branch noise (`Swiggy Limited Banga` → `Swiggy`) so the refund nets against the original spend merchant.
- **`detectIsCard`** — recognise `credit card` + `card XX<digits>` as a card transaction.

Spend-side CREDIT typing + available-limit capture were already correct; unchanged.

## Tests
Tightened the existing refund test to assert `merchant = "Swiggy"`, `accountLast4 = "1234"`, `isFromCard = true` (spend case already asserts `CREDIT` + `creditLimit = 48750.00`). All 28 IndusInd tests pass; full `:parser-core:jvmTest` green via `./init.sh parser`.

## Out of scope (app-side, left for owner)
- Whether `Avl Lmt` (remaining limit) should feed the model's `creditLimit - balance` derivation.
- Monthly credit-card refund netting in `HomeViewModel` / `DEDUCT_SPENT` for the CREDIT bucket.

### DIFF
--- a/parser-core/src/main/kotlin/com/pennywiseai/parser/core/bank/IndusIndBankParser.kt
+++ b/parser-core/src/main/kotlin/com/pennywiseai/parser/core/bank/IndusIndBankParser.kt
@@ -65,10 +65,9 @@ class IndusIndBankParser : BaseIndianBankParser() {
         // Credit-card refund SMS say "credited to your IndusInd Bank Credit Card XX1234 ...
         // adjusted against the outstanding on your card account". The trailing "card account"
         // phrase makes the base detectIsCard bail out as an account (non-card) txn, so
-        // recognise the credit card explicitly here. (#486)
-        if (lower.contains("credit card") &&
-            Regex("""card\s+x{2}\d""", RegexOption.IGNORE_CASE).containsMatchIn(message)
-        ) {
+        // recognise the credit card explicitly here. Mask is 2+ X's so XX1234 and
+        // XXXX1234 both match. (#486)
+        if (lower.contains("credit card") && CARD_MASK_PATTERN.containsMatchIn(message)) {
             return true
         }
         return super.detectIsCard(message)
@@ -181,16 +180,9 @@ class IndusIndBankParser : BaseIndianBankParser() {
         // Credit-card refund: "refund of INR <amt> from <Merchant> has been credited ..."
         // IndusInd appends the legal entity + branch/location (e.g. "Swiggy Limited Banga");
         // keep just the brand so the refund nets against the original spend merchant. (#486)
-        val refundPattern = Regex(
-            """refund\s+of\s+(?:INR|Rs\.?|₹)\s*[0-9,]+(?:\.\d{2})?\s+from\s+(.+?)\s+has\s+been\s+credited""",
-            RegexOption.IGNORE_CASE
-        )
-        refundPattern.find(message)?.let { match ->
+        REFUND_MERCHANT_PATTERN.find(message)?.let { match ->
             var m = match.groupValues[1].trim()
-            m = m.replace(
-                Regex("""\s+(?:Limited|Ltd\.?|Pvt\.?|Private).*$""", RegexOption.IGNORE_CASE),
-                ""
-            )
+            m = m.replace(LEGAL_ENTITY_SUFFIX_PATTERN, "")
             if (m.isNotEmpty()) return cleanMerchantName(m)
         }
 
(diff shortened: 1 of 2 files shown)


sarim2000:
Addressed the 4/5 review:
- Hoisted `CARD_MASK_PATTERN`, `REFUND_MERCHANT_PATTERN`, and `LEGAL_ENTITY_SUFFIX_PATTERN` into a private companion object (compiled once, not per-SMS).
- Relaxed the card mask `x{2}\d` → `x{2,}\d` so a 4-X mask (`XXXX1234`) is still detected as a card txn instead of falling through to the base class.
- Added a refund test with an `XXXX5678` mask.

`:parser-core:jvmTest` green.

PR #617, opened 2026-07-18
fix(parser): Indian Bank "Sent Rs." UPI-debit was dropped (returned null)

sarim2000:
## Problem
Indian Bank's newer UPI-debit SMS format is **silently dropped** — `BankParserFactory.parse(...)` returns `null`, so the transaction is never recorded (real data loss). Reported from a live SMS:

> Sent Rs.440.00 from A/c *4512 on 17-07-26 to NARMADA FOODS.RRN 213416112187.Avl Bal Rs.4585.65.Not you? SMS BLOCK to 9289592895-Indian Bank

**Root cause:** the base `BankParser.isTransactionMessage` only recognises verbs like `debited/credited/withdrawn` — not `Sent` — so the message fails the is-this-a-transaction gate before any extraction runs. (The "Avl Bal" + no-known-verb combination would also trip the balance-update-only path.)

## Fix
All changes are **bank-local overrides on `IndianBankParser`** — no edits to `BaseIndianBankParser`/`BankParser`, so no other bank that extends the base is affected:
- `isTransactionMessage` — also accept `"sent rs"` so `parse()` no longer drops it.
- `isBalanceUpdateNotification` — return `false` for `"sent rs"` so a debit carrying `Avl Bal` isn't misclassified as a balance-update-only notice.
- `extractAmount` — add `Sent\s+Rs\.?\s*<amt>` pattern.
- `extractTransactionType` — `"sent rs"` → `EXPENSE`.
- `extractMerchant` — `to <merchant>.RRN` → merchant (stops at the period before RRN).
- `extractReference` — `RRN\s+<digits>`.

## Tests
Added a `TestIndianBankParser` case asserting `type=EXPENSE`, `amount=440.00`, `merchant="NARMADA FOODS"`, `accountLast4="4512"`, `reference="213416112187"`, `balance=4585.65`. Full `:parser-core:jvmTest` green via `./init.sh parser`.

### DIFF
--- a/parser-core/src/main/kotlin/com/pennywiseai/parser/core/bank/IndianBankParser.kt
+++ b/parser-core/src/main/kotlin/com/pennywiseai/parser/core/bank/IndianBankParser.kt
@@ -190,12 +190,6 @@ class IndianBankParser : BaseIndianBankParser() {
     }
 
     override fun extractReference(message: String): String? {
-        // Pattern 0: RRN 213416112187 (newer UPI-debit format)
-        val rrnPattern = Regex("""RRN\s+(\d+)""", RegexOption.IGNORE_CASE)
-        rrnPattern.find(message)?.let { match ->
-            return match.groupValues[1]
-        }
-
         // Pattern 1: UPI:515314436916
         val upiRefPattern = Regex("""UPI:(\d+)""", RegexOption.IGNORE_CASE)
         upiRefPattern.find(message)?.let { match ->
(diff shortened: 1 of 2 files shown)


sarim2000:
Addressed the 4/5 review:
- **RRN priority** — moved the RRN pattern *below* the `UPI:` / `UPI Ref no` patterns in `extractReference`, so a message carrying both keeps preferring the UPI ref; the "Sent Rs." format carries only an RRN so it's still captured.
- **Amount coverage** — added two `Sent Rs.` cases: comma-grouped (`Rs.1,500.00`) and whole-number (`Rs.500`), exercising the comma-strip and optional-decimal paths.

`:parser-core:jvmTest` green.

PR #623, opened 2026-07-19
feat(add): manual account-to-account Transfer with From/To accounts (#622)

### LINKED ISSUE #622: [BUG] Transfer type in Add Transaction shows a single account picker (no From/To)
Reported on Discord. In the Add Transaction form, selecting **Transfer** shows the exact same single **Select Account** picker as Expense — there's no way to pick a *From* and *To* account, so a manual transfer can't actually be recorded as moving money between two accounts.

### Root cause
The Transfer chip was added to the type selector, but the form was never given a two-account UI for it:
- `TransactionTabContent.kt` renders one `selectedAccount` picker unconditionally (no branch on `transactionType`).
- `AddViewModel` holds only a single `selectedAccount`; `AddTransactionUseCase.execute(...)` accepts only one `bankName`/`accountLast4`.
- `TransactionEntity` already has `fromAccount`/`toAccount` columns (used by the SMS parser for auto-detected transfers) but the manual Add flow never populates them.

### Fix
When `transactionType == TRANSFER`:
1. Show **From account** + **To account** pickers (hide Merchant; keep Category optional/default).
2. Populate `fromAccount`/`toAccount` on the entity.
3. Move the money: decrement the source and increment the destination in one DB transaction (reusing the existing manual-recompute / SMS-delta balance model).

Same-currency transfers only for the first cut (no cross-currency conversion — consistent with the app's never-sum-across-currencies rule).

sarim2000:
Closes #622.

## Problem
Selecting **Transfer** in Add Transaction showed the exact same single "Select Account" picker as Expense (reported on Discord) — there was no way to record money moving between two accounts. The Transfer chip existed, but the form/ViewModel/use-case never supported a From→To pair.

## What this does
When `transactionType == TRANSFER`:
- **UI** (`TransactionTabContent`): hides Merchant, shows **From account** + **To account** pickers (a shared account bottom-sheet routes selection to whichever is open via an `AccountPickerTarget`). All other types unchanged.
- **ViewModel**: adds `toAccount` state; `selectedAccount` is the From account. `isValid`/`saveTransaction` become transfer-aware — no merchant required; requires two **distinct** accounts of the **same currency** (the From account fixes the entry currency). Cross-currency is rejected (we never sum across currencies).
- **Use case / repo**: populates `fromAccount`/`toAccount` and moves the money in one atomic Room transaction — `AccountBalanceRepository.insertTransferWithBalance` debits the FROM leg and credits the TO leg, using manual-recompute for cash/manual accounts and a signed delta for SMS-tracked accounts.

## Bug caught during emulator testing (and fixed)
The first on-device run debited the **wrong** account: a transfer from Kotak ••9999 hit HDFC ••9999 instead, because the SMS-delta leg resolved accounts by **last4 alone**. Fixed by resolving the delta by **(bankName, last4)** (`insertBalanceDelta`) — the manual-add path knows the full bank for both legs. Re-verified: Kotak ••9999 −500, Kotak ••7777 +500, HDFC ••9999 untouched.

## Verification (emulator)
- Transfer form shows From/To, Merchant hidden, Save gated until valid.
- Saved transfer is recorded as a transfer — **excluded from "Spent this month"**, shown under "Money in motion".
- From −amount and To +amount both move correctly, including the shared-last4 case.
- `./init.sh app` green.

## Known limitation (follow-up, pre-existing)
Editing/deleting a transfer reverts via the existing last4-only `applyTransferBalanceShift` (the persisted `from_account`/`to_account` columns store only last4, so the bank isn't available at revert time). A shared-last4 **edit/delete** can therefore mis-revert — the same limitation that already affects SMS-detected transfers. Out of scope here; would need the leg columns to carry full account identity.

Same-currency transfers only for this first cut.

### DIFF
--- a/app/src/main/java/com/pennywiseai/tracker/data/repository/AccountBalanceRepository.kt
+++ b/app/src/main/java/com/pennywiseai/tracker/data/repository/AccountBalanceRepository.kt
@@ -318,6 +318,58 @@ class AccountBalanceRepository @Inject constructor(
         rowId
     }
 
+    /**
+     * Atomically insert a manual TRANSFER [transaction] and move BOTH legs'
+     * balances as one Room transaction, so a crash can never leave the money
+     * debited from one account but never credited to the other (or vice-versa).
+     * The [transaction] must already carry `transactionType = TRANSFER`,
+     * `fromAccount = fromLast4` and `toAccount = toLast4`.
+     *
+     * The two legs use different balance mechanics, mirroring the rest of this
+     * class:
+     *  - Manual/cash accounts have a *derived* balance (opening + Σtxns, with
+     *    TRANSFER handled by `from_account`/`to_account` in [signedTransactionSum]).
+     *    Once the row is inserted, [recomputeManualBalance] re-derives the figure —
+     *    no explicit delta needed. Opening anchors for both legs are pinned from
+     *    the PRE-insert snapshot via [ensureManualOpening] before the insert, so
+     *    the recompute includes this new transfer exactly once.
+     *  - SMS-tracked accounts have a bank-reported balance, so we snapshot a signed
+     *    "TRANSACTION" delta off the latest row via [insertBalanceDeltaByLast4]
+     *    (−amount on the FROM leg, +amount on the TO leg). No-op if that account
+     *    has no balance row yet.
+     *
+     * Returns the new row id (-1 on a dedup-conflict insert, in which case no
+     * balance is moved).
+     */
+    suspend fun insertTransferWithBalance(
+        transaction: TransactionEntity,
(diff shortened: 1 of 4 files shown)


sarim2000:
Addressed the 3/5 review:
- **Dedup hash** — `generateTransferHash` now includes both **bank names** (not just last4), so two distinct transfers sharing a last4 (Kotak ••9999 vs HDFC ••9999 → same destination/amount/second) can't hash-collide and silently no-op the second insert. Now consistent with the bank-aware `(bankName, last4)` resolution in the balance move.
- **Stale toAccount** — `updateTransactionType` clears `toAccount` when leaving TRANSFER.

`./init.sh app` green.

PR #627, opened 2026-07-19
feat(import): CSV transaction importer (#299)

### LINKED ISSUE #299: [FEATURE] Import CSV from others App to add historical txns?
## 💡 Feature Description
Axio app was the older app that does what pennywise does; reading SMS to parse txns. I want to move away from it, as it is close sourced. If we can import the csv from axio directly to this app, it would be really helpful to carry over our history and past txns. Having a csv import with a standardized format can help many leave their apps and move to pennywise instead.

## 🎯 Problem it Solves
Allows user from another similar app to easily onboard to Pennywise. Also, we can have a standard csv import format, which can be used to import any set of txns into pennywise, not just from axio.

## 🔧 Proposed Solution
Allowing user to import csv provided from axio's export, or any standard csv format can work.

## 📱 Alternatives Considered
Maybe adding txns from bank's statement pdfs? But this would be longer and tougher to implement.

## 💬 Additional Context
For instance, the Firefly III app, for budget and transaction monitoring, it allows us to import csvs or export it.

sarim2000:
Closes #299.

## What
Lets users import historical transactions from a CSV in **PennyWise's own export format** — the 12 columns `CsvExporter` writes (`Date, Time, Merchant, Category, Type, Amount, Currency, Bank, Account, Balance After, Description, SMS Body`). So an export round-trips, and someone migrating from another app (axio, etc.) can carry their history over by reformatting to that standard layout.

MVP scope is deliberately the round-trip format — no per-app column-mapping UI (that's a possible fast-follow).

## How
- **`CsvTransactionImporter`** (pure/testable — takes a `java.io.Reader`): OpenCSV parse, **header-name** column matching (tolerant of column reorder / extra columns), requires Date/Amount/Type, sensible defaults for the rest (`bank`→"Imported", `category`→"Others", `currency`→"INR"). Per-row failures (bad amount/date/unknown type) are **counted, not thrown**. Currency is set per-row from the CSV — nothing summed across currencies.
- **`ImportCsvUseCase`**: opens the picked `Uri`, dedups by a **deterministic content hash** (against the DB via `getTransactionByHash` and within the same file), inserts the rest; returns imported / skippedDuplicate / failed.
- **Settings**: an "Import Transactions (CSV)" entry (GetContent picker) beside "Import Data"; result surfaced via the existing import/export snackbar. **Free** — it's a migration/onboarding driver, not Pro-gated.
- **Test**: `CsvTransactionImporterTest` (5 cases, incl. a bad-amount row counted as failed).

## Verification (emulator)
Pushed a 2-row CSV (Expense + Income) and imported via the new entry → **"Imported 2 transactions, skipped 0 duplicates"**; both appeared with correct amount/type/category/description/currency. Re-imported the same file → **"Imported 0 transactions, skipped 2 duplicates"** (dedup works). `./init.sh app` green.

## Note / possible follow-up
Imports populate history & analytics but **don't retroactively recompute account balances** (historical imports shouldn't rewrite *current* balances); rows without a Bank land in an "Imported" pseudo-account. If balance-affecting imports are wanted for manual/cash accounts, that's a separate decision.

### DIFF
--- a/app/src/main/java/com/pennywiseai/tracker/data/csv/CsvTransactionImporter.kt
+++ b/app/src/main/java/com/pennywiseai/tracker/data/csv/CsvTransactionImporter.kt
@@ -96,7 +96,9 @@ class CsvTransactionImporter @Inject constructor() {
                 return ImportParseResult(
                     transactions = emptyList(),
                     failedCount = 0,
-                    failureReasons = listOf("Missing required column(s): $missing")
+                    failureReasons = listOf(
+                        "This doesn't look like a PennyWise CSV export — missing column(s): $missing"
+                    )
                 )
             }
 
--- a/app/src/main/java/com/pennywiseai/tracker/data/csv/CsvTransactionImporter.kt
+++ b/app/src/main/java/com/pennywiseai/tracker/data/csv/CsvTransactionImporter.kt
@@ -1,0 +1,278 @@
+package com.pennywiseai.tracker.data.csv
+
+import com.opencsv.CSVReader
+import com.pennywiseai.tracker.data.database.entity.TransactionEntity
+import com.pennywiseai.tracker.data.database.entity.TransactionType
+import java.io.Reader
+import java.math.BigDecimal
+import java.security.MessageDigest
+import java.time.LocalDate
+import java.time.LocalDateTime
+import java.time.LocalTime
+import java.time.format.DateTimeFormatter
+import javax.inject.Inject
+import javax.inject.Singleton
+
+/**
+ * Parses PennyWise's own CSV export format (the 12 columns [CsvExporter] writes)
+ * back into [TransactionEntity] rows, so historical data can be re-imported and
+ * an export can round-trip.
+ *
+ * Pure/testable: takes a [Reader] (no Android `Uri`), so it can be exercised from
+ * a plain JUnit test with a `StringReader`. The Android side (opening the picked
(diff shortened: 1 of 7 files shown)


sarim2000:
Addressed the 4/5 review:
- **Per-row DB queries** → now a single **batch** lookup (`TransactionDao.getExistingHashes` with an `IN` clause) for the whole file, so a large CSV no longer issues N `getTransactionByHash` queries.
- **Amount sign** → non-positive amounts (`signum() <= 0`) are now **rejected as a failed row** (amounts are always positive in the format; direction is carried by Type), so a hand-crafted signed-debit CSV can't silently corrupt analytics. Added a test for negative + zero.

`./init.sh app` green.

sarim2000:
Addressed the re-review:
- **P1 (IN-clause crash on 1000+ rows)** — `getExistingHashes` now **chunks at 900** in the repository (under SQLite's 999 bind-variable limit; Room doesn't auto-chunk), so a large CSV can't crash the import.
- **P2 (MIME / confusing error)** — kept `"*/*"` (matches the sibling Import Data, and file managers frequently tag CSVs as `application/octet-stream`, so `text/csv` would hide legit files), but a wrong-file pick now surfaces an explicit **"This doesn't look like a PennyWise CSV export — missing column(s): …"** error instead of a silent "Imported 0".

`./init.sh app` green.

PR #628, opened 2026-07-19
feat(fdroid): "Support development" UPI tip jar on F-Droid builds (#503)

### LINKED ISSUE #503: [FEATURE] Add 'Support development' section for F-Droid/APK users (UPI + Ko-fi)
## Goal
APK/F-Droid users get the app fully free with **no way to support development** (no Play billing in the `fdroid` flavor). Add a subtle "Support development" section for those builds. Play users keep the Pro upgrade instead.

## Decision (after research)
**Two rails:** **UPI** (India — most of our base) + **Ko-fi** (global).

| Option | Verdict |
|---|---|
| **UPI** | ✅ Primary. Native (`upi://pay` intent, no browser), free, instant, zero setup. Covers the largely-Indian base. |
| **Ko-fi** | ✅ Global rail. 0% fee, pays out to India via PayPal, available for Asian creators. |
| ~~Buy Me a Coffee~~ | ❌ Drop — Stripe-only payouts, **India not supported** (can't withdraw). |
| GitHub Sponsors | ⏳ Optional later. Verified India payout but needs Stripe-KYC enrollment. For in-app it's redundant with Ko-fi; its real value is the repo Sponsor button (just `.github/FUNDING.yml`, no app code). |

## Scope / tasks
- [ ] **Flavor-gate to F-Droid/APK only** — keep the donate section OUT of the `standard` (Play) build (Google's external-payment policy; Play has Pro). Gate via the build flavor, same pattern as `standardImplementation`/app-update.
- [ ] **Settings/About "❤️ Support development" section** (subtle, on-brand yellow tonal — matches our Pro-prompt style; no nag/popup):
  - **UPI** row → `upi://pay?pa=<UPI_ID>&pn=<Payee Name>&cu=INR` intent. Label as the native/India option. Handle "no UPI app installed" gracefully.
  - **Ko-fi** row → opens `https://ko-fi.com/<username>` in browser. Label e.g. "Support from outside India".
- [ ] **fdroiddata Donate field** → add the Ko-fi (or Sponsors) URL so the F-Droid listing shows a Donate button.
- [ ] (Optional) **`.github/FUNDING.yml`** with Ko-fi + (later) GitHub Sponsors → repo Sponsor button.
- [ ] There's an existing unmerged `feat/github-sponsors-link` branch (Settings + repo Sponsor button) — reuse/adapt its UI scaffolding.

## Prerequisites (manual, by maintainer)
- [ ] Create a **dedicated UPI ID** for donations (separate from personal — UPI ID is receive-only/safe to expose, but isolate name + collect-request spam). Note the payee name shown to payers.
- [ ] Create **Ko-fi** account + connect PayPal (India payout).

## Notes
- UPI ID is safe to expose publicly (receive-only; can't be debited without our PIN). Use a dedicated handle.
- Reference: the Cashiro fork uses the same shape (UPI + a coffee link) in its AboutScreen.
- UPI is India-only by design; Ko-fi covers global.

sarim2000:
Closes #503.

## Why
F-Droid builds hardcode `isPro = true` (no Play billing allowed by F-Droid policy), so those users get everything free but had **no way to support development**. The Pro upsell is meaningless there. This adds a subtle "Support development" tip jar on the **fdroid flavor only**; Play builds keep the Pro upgrade.

## What
- Gated on `BuildConfig.FLAVOR == "fdroid"` — the top-of-Settings slot shows **Support development** on F-Droid, **PennyWise Pro** on Play.
- Tapping opens a small dialog: warm one-liner, the developer's **UPI VPA** shown with a **copy** button, and **Pay via UPI** → fires `upi://pay?pa=…&pn=PennyWise&cu=INR&tn=PennyWise support` (opens the user's UPI app; amount left blank for them). Falls back to a "copy the UPI ID" hint via toast if no UPI app is installed.
- VPA + payee name are `translatable="false"` string resources (public donation address).

**Rails:** UPI is the primary rail — largely-Indian base, zero fees, native intent, no proprietary SDK (F-Droid-safe). A global/Ko-fi link can be added later once payouts are set up (per #503's two-rail plan).

## Verification (emulator)
F-Droid has no x86 build, so I previewed the fdroid path by temporarily forcing the flag on the standard build (then reverted):
- Support section renders in place of Pro; dialog + copy + Pay-via-UPI work; the `upi://pay` intent fires (chooser opens — no UPI app on the emulator, which is the fallback case).
- With the flag reverted, the **standard** flavor still shows the Pro upsell and no Support section.
- `./init.sh app` green.

### DIFF
--- a/app/src/main/java/com/pennywiseai/tracker/ui/screens/settings/SettingsScreen.kt
+++ b/app/src/main/java/com/pennywiseai/tracker/ui/screens/settings/SettingsScreen.kt
@@ -31,7 +31,12 @@ import androidx.compose.ui.Modifier
 import androidx.compose.ui.draw.clip
 import androidx.compose.ui.graphics.Color
 import androidx.compose.ui.graphics.vector.ImageVector
+import android.widget.Toast
+import androidx.compose.ui.platform.LocalClipboardManager
 import androidx.compose.ui.platform.LocalContext
+import androidx.compose.ui.res.stringResource
+import androidx.compose.ui.text.AnnotatedString
+import com.pennywiseai.tracker.R
 import androidx.compose.ui.text.font.FontWeight
 import androidx.compose.ui.text.style.TextAlign
 import androidx.compose.ui.unit.dp
@@ -121,6 +126,10 @@ fun SettingsScreen(
     val scheduledFolderBackupLastTimestamp by settingsViewModel.scheduledFolderBackupLastTimestamp.collectAsStateWithLifecycle(initialValue = null)
     val requestFolderPicker by settingsViewModel.requestFolderPicker.collectAsStateWithLifecycle()
     var showUpgradeSheet by remember { mutableStateOf(false) }
+    var showSupportDialog by remember { mutableStateOf(false) }
+    // F-Droid builds have no Play billing, so they show a "Support development"
+    // tip jar instead of the (un-buyable) Pro upsell. Play builds keep Pro.
+    val isFdroidBuild = com.pennywiseai.tracker.BuildConfig.FLAVOR == "fdroid"
     // Launches the runtime permission request. If granted, we flip the
     // preference on; if denied, leave the switch off so the user can try
     // again without us silently turning the feature on later.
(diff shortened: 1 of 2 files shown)


sarim2000:
Addressed the 4/5 (both cosmetic):
- **Hardcoded strings** → the Support-dev section/dialog user-facing strings (title, subtitle, body, button labels, copy content-desc, toasts) are now `strings.xml` resources.
- **Fully-qualified names** → `launchUpiPayment` uses imported `Context`/`Uri`/`ActivityNotFoundException`, and takes the no-UPI-app message as a param resolved via `stringResource`.

No behavior change; `./init.sh app` green.

sarim2000:
Good catch — fixed the real one: `launchUpiPayment` now returns whether a UPI app actually opened, and the dialog **only dismisses on success**. If no UPI app is installed, the dialog stays up so the copy-the-VPA fallback the toast points to is still on screen. (The other two inline comments are stale re-anchors — the summary confirms strings/URI are already correct.)

PR #632, opened 2026-07-23
feat(export): carry analytics custom range into drill-down + group CSV export

sarim2000:
## What

Two gaps in "export exactly what I'm looking at":

### 1. Analytics custom range now carries into the drill-down
Tapping a merchant / category / tag in Analytics while a **custom date range** was active previously landed on the Transactions screen reset to **This Month** (the nav handoff only mapped THIS_MONTH / LAST_MONTH / CURRENT_FY / ALL). The custom range now travels with the navigation as epoch-day route args and is applied on arrival — so the existing Export FAB exports exactly the slice that was on screen in Analytics.

- `AnalyticsScreen`: passes `customDateRange` (epoch days) through `onNavigateToTransactions` when the period is CUSTOM
- `MainScreen`: `startDate`/`endDate` route args on the transactions destination
- `TransactionsViewModel`: shared `applyNavigatedPeriod()` used by both `applyInitialFilters` and `applyNavigationFilters`; `NavigationParams` includes the range so the dedup check stays correct

### 2. Group-wise CSV export
The transaction-group detail screen's ⋮ menu gains **Export CSV** (hidden while the group is empty), reusing `ExportTransactionsDialog` on the group's linked transactions — free-tier row limit / Pro behavior identical to the Transactions screen export.

## Verification
- `./init.sh app` green (build + unit tests)
- On-emulator E2E:
  - Custom range **Jul 19–23** → tap Zomato in Top Merchants → Transactions shows the **"Jul 19 - Jul 23"** chip + only the 2 in-range Zomato txns → export dialog offers exactly 2 transactions
  - Range filtering proven exclusive: switching the range to Jul 20–22 correctly hides Jul 23 transactions
  - Created a group with 2 txns → ⋮ → Export CSV → "Export Complete! 2 transactions" → pulled the CSV and verified both rows in the standard 12-column format

### DIFF
--- a/app/src/main/java/com/pennywiseai/tracker/presentation/groups/TransactionGroupDetailScreen.kt
+++ b/app/src/main/java/com/pennywiseai/tracker/presentation/groups/TransactionGroupDetailScreen.kt
@@ -25,6 +25,7 @@ import androidx.hilt.navigation.compose.hiltViewModel
 import androidx.lifecycle.compose.collectAsStateWithLifecycle
 import com.pennywiseai.tracker.data.database.entity.TransactionEntity
 import com.pennywiseai.tracker.data.database.entity.TransactionType
+import com.pennywiseai.tracker.presentation.transactions.ExportTransactionsDialog
 import com.pennywiseai.tracker.ui.components.CustomTitleTopAppBar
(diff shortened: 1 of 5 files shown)


sarim2000:
@greptile review

PR #643, opened 2026-07-26
Bank landing pages, Play listing rewrite, and a parser-history content miner

sarim2000:
Four related pieces of distribution work. Each stands alone; the middle two are coupled (the website renders what the parser generates).

## 1. Play Store listing

The listing advertised **"90+ banks across 16 countries"** while the registry had grown to **141 across 23** — we were selling 64% of what the app does.

| | Before | After |
|---|---|---|
| Title | `PennyWise AI - Expense Tracker` | `PennyWise: SMS Expense Tracker` (30/30) |
| Short | no UPI, no "SMS" | `Auto-track UPI & bank SMS…` (78/80) |
| Coverage | 90+ banks / 16 countries | **141 / 23** |
| SMS permission | unaddressed | second section, ahead of features |

Dropping "AI" from the title buys the "SMS" keyword; it was brand-only and won no queries. One-word revert if you disagree.

The stale count is now structurally impossible — `SupportedBanksDocTest` already regenerated the README from the live registry, so it owns the listing's coverage sentence too. Play descriptions can't carry HTML comment markers, so the sentence itself is the marker. **Tamper-tested:** flipping the numbers fails the build with a pointer to the fix script.

Validated against Google Play with `upload-play.sh listing --dry-run` (`validate_only: true`, both locales, no errors). **Nothing published — that's your call.**

## 2. A real parse demo per bank

`BankSamplesDocTest` harvests one already-anonymised message per bank from the existing test corpus, runs it through the live `BankParserFactory`, and emits `bank-samples.json`. Both halves are real, so a landing page can't demo an extraction the app wouldn't make.

**92 of 141 banks** yield a clean demo. Samples are quality-gated — a candidate is skipped when the parse is degenerate. That picks *which* real example to show and never edits one.

> **⚠️ Worth fixing separately:** the gate exists because those are real bugs. HDFC and SBI return `"A/c XX1234"` as the merchant on some messages, and Telebirr's `accountLast4` comes back as `"Name"` (so it has no clean sample). This lines up with the miner's finding that merchant extraction is 21% of all parser fixes.

## 3. 165 landing pages

The site had two routes and no sitemap. Now: 141 bank pages, 23 country pages, an index, `sitemap.xml`, `robots.txt` — all generated from the same catalogue, so we can never publish a page for a bank we don't parse.

The centrepiece is the real parse: the message with the exact substrings the parser consumed tinted in place, and the fields it produced underneath in matching colours. **The colour is the mapping.** That's what keeps these off Google's thin-doorway radar.

`ParseViews` carried its own copy of the header and ~50 lines of duplicate CSS — which is why nav additions never reached the site's most-visited page. It now shares the chrome.

**SEO measured across all 166 pages, not assumed:** `lang="en"`, one `h1` each, self-consistent canonicals, unique titles ≤60, unique descriptions ≤160, FAQPage + BreadcrumbList JSON-LD, sitemap with no dead URLs (asserted in tests).

> **Known, not fixed:** bank pages share **~0.51 median 6-gram Jaccard similarity** — the real indexing risk for programmatic pages. I'd ship the 92 with demos first rather than all 141 at once. The webfont link is also render-blocking; worth self-hosting the two subsets before launch.

## 4. Content miner

`node scripts/content-mine/index.mjs` — pure git + node, no deps. `--draft` adds a post draft via the Agent SDK.

**The obvious framing doesn't survive the data.** First commit is 2025-08-01, so most parser commits are initial coverage, not format churn — exactly one commit says a bank changed its messaging. "N banks changed their SMS format this year" would not be defensible, so the prompt is fenced to the mined numbers and forbids that claim.

What *is* supported: **43% of parsers needed a fix >30 days after shipping** (ICICI nine times); failure modes are card/account confusion 21%, merchant extraction 21%.

## Verification

- `./init.sh` — passes (full gate, rebased on `main`)
- `pennywise-web` — 8/8 tests, including one that walks every sitemap URL and fails on any 404
- Ran locally against a throwaway Postgres; all 167 routes checked, pages reviewed at desktop and mobile widths

Two bugs caught by looking at rendered output rather than tests: the primary CTA was **white-on-white and invisible** on all 165 pages (`.prose a` out-specifying `.btn-primary`), and `"Reference or UTR number"` was showing on the Tanzanian M-Pesa page (UTR is Indian NEFT/RTGS; now region-aware).

## Not included

Screenshots 1 and 2 in the listing still show a masked balance, an empty-state Budgets card, and a FAB covering a transaction amount. Needs an emulator re-shoot with seeded data — separate PR.


sarim2000:
All three findings were real. Fixed in 9104796.

## 1. Cross-bank sample pairing — confirmed, and worse than described

The pairing keyed off proximity ("first `sender =` after this `message =`"). That breaks whenever a constructor lists sender first — `SimpleTestCase` does, `ParserTestCase` doesn't. In a file covering several banks the scan walked past the end of the block and grabbed the **next bank's** sender.

Concrete case: `ThailandBankParsersTest.kt` has a `SimpleTestCase` for UOB with `sender` before `message`, so the UOB message got filed under **CIMB Thai** and published on `/banks/cimb-thai`.

I diffed the regenerated file against what was on the branch — **6 of 92 pages were showing another bank's message** (CIMB Thai, KTC, Krungsri, Siam Commercial Bank, TTB, Tigo Pesa — all from multi-bank test files).

Now scans balanced parens and takes both fields from the same test-case block, so pairing is correct by construction rather than by luck. Added a regression test built from the exact shape that broke.

`CIMB Thai` → `CIMB: Transfer received 6,000.00 THB…` · `UOB Thailand` → `UOB: Card transaction 3,200.00 THB…`

## 2. Base classes counted as parsers — confirmed

The denylist had already rotted: `BaseIranianBankParser` and `BaseThailandBankParser` were being counted. 143 reported vs 141 in the registry.

Replaced the hand-maintained list with a derived check — a file declaring an `abstract`/`sealed` class of its own name is scaffolding, not a bank — so it can't rot again. Now reports **141, matching the registry exactly**. The headline stat moves from 43% → 44%.

## 3. Google Fonts — confirmed, and the fix went further

You're right, and it's worse than a generic privacy nit given what this site is advertising: the marketing pages for an app whose whole claim is "nothing leaves your device" were handing every visitor's IP, User-Agent and Referer to Google before first paint.

Both families are now **self-hosted** — 55KB of latin-subset woff2, SIL OFL, attribution in `static/fonts/LICENSE.md`. That also removes the render-blocking cross-origin round trip I'd flagged separately as a Core Web Vitals cost, so it closes both.

While in there: **htmx was loading on all 167 pages** but is only used by the parser and feedback pages — this PR had expanded that request from 2 pages to 167. Now opt-in per page.

**Verified by driving a real browser and recording every non-localhost request:**

```
external hosts contacted: NONE
font faces: Archivo 600 800 loaded | IBM Plex Mono 400 loaded | IBM Plex Mono 500 loaded
```

Landing pages now issue **zero** third-party requests. The parser page still loads htmx from unpkg — pre-existing, and it genuinely uses it; happy to self-host that too if you want it gone.

## Verification

`./init.sh` passes · web tests 8/8 (7 route + 1 existing) · rendering confirmed unchanged after the font swap.

## Still open from the original description

Unchanged by this round, and still worth your call before merge: bank pages share **~0.51 median 6-gram Jaccard similarity**, so I'd still ship the 92 with demos before all 141. And the parser bugs the generator surfaced (HDFC/SBI merchant → `"A/c XX1234"`, Telebirr `accountLast4` → `"Name"`) deserve their own fix — the quality gate hides them from the marketing pages but they're still live in the app.

PR #671, opened 2026-08-12
feat(security): pin chat model to versioned URL + verify SHA-256 before load

sarim2000:
## Problem

The on-device chat model (`Qwen2.5-1.5B-Instruct-q8-ekv4096.litertlm`) was downloaded from a **bare R2 bucket path** with no version pinning and **no integrity check**:

- `DownloadManager` fetched the file; on `STATUS_SUCCESSFUL` it was marked `READY`.
- `isModelDownloaded()` accepted it on **file size alone** (≥95% of expected).
- `LlmRepository.ensureConversation()` loaded whatever file was at that path into MediaPipe, unconditionally.

So if the served bytes ever changed in place — host account handover, a CDN issue, or an accidental/hostile overwrite — the app would **silently load a swapped model** inside a finance app, with no version bump to make it noticeable. (HTTPS already covers the in-transit case; this is about the object changing at rest.)

Reported privately by an external researcher as a hardening suggestion (not an exploit).

## Change

1. **Versioned, immutable URL** — `MODEL_URL` is now built from `models/$MODEL_VERSION/$MODEL_FILE_NAME` instead of the bucket root, so a given URL always resolves to the same bytes. Publishing a new model means a new versioned key + a hash bump, never an in-place overwrite.
2. **Pinned SHA-256** — `MODEL_SHA256` ships in the app; the downloaded file is hashed and compared before it is ever loaded.
3. **Verify-before-load** — `LlmRepository.ensureConversation()` calls `verifyModelIntegrity()` and refuses to load (deletes the file) on mismatch.
4. **Verify-before-READY** — `ChatViewModel` verifies a freshly-downloaded file before flipping to `READY`; a bad file is deleted and surfaced as an error.

`verifyModelIntegrity()` streams the SHA-256 (64 KB buffer), caches the result per process by file length so we don't re-hash 1.5 GB on every message, and treats a blank hash as "verification intentionally disabled" (logged warning).

## Infra

The model has been copied server-side in R2 to the new versioned key `models/v1/…`. The public URL was downloaded and confirmed to hash to the pinned value; the old root object is left intact for backward compatibility with already-shipped builds.

## Verification

- `./init.sh app` — green (compile + unit tests).
- Public `models/v1/` URL: `HTTP 200`, `Content-Length` 1,597,931,520, SHA-256 == pinned `faa60663…1a2dc9`.
- Recommended follow-up: on-device pass of the download → verify → ready path before release.

### DIFF
--- a/app/src/main/java/com/pennywiseai/tracker/core/Constants.kt
+++ b/app/src/main/java/com/pennywiseai/tracker/core/Constants.kt
@@ -72,10 +72,38 @@ object Constants {
     
     /**
      * LLM Model Configuration
+     *
+     * The chat model is distributed from a **versioned, immutable object key**
+     * and verified against a pinned SHA-256 before it is ever loaded. To publish
+     * a new model: upload it under a fresh [MODEL_VERSION] key (never overwrite an
+     * existing version in place), then bump [MODEL_VERSION] and [MODEL_SHA256]
+     * together. This makes a silently-swapped model impossible to ship.
      */
     object ModelDownload {
-        const val MODEL_URL = "https://pub-fcfb3ffddb184540a758a7fe68249908.r2.dev/Qwen2.5-1.5B-Instruct-q8-ekv4096.litertlm"
         const val MODEL_FILE_NAME = "Qwen2.5-1.5B-Instruct-q8-ekv4096.litertlm"
+
+        /** Bump when a new model build is published under a new versioned key. */
+        const val MODEL_VERSION = "v1"
+
+        private const val R2_PUBLIC_BASE =
+            "https://pub-fcfb3ffddb184540a758a7fe68249908.r2.dev"
+
+        /**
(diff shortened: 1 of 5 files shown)


desledishant10:
Thank you, this is a great fix and a fast one. I really appreciate you taking it seriously and going all the way with it rather than just papering over it.

A few things I especially liked: re-hashing at the load boundary in ensureConversation(), not just after download, closes the time-of-check-to-time-of-use gap cleanly, and moving to the versioned models/v1 key kills the in-place-overwrite problem at its root. Nice touch testing the reject path on-device too, and good call resolving the Greptile P1s before merge, the stale length-cache one in particular was subtle.

I'll credit you by name in the capstone writeup, Sarim (@sarim2000), as you offered. Happy to share the short section once it's public if that's useful. Thanks again, and for building PennyWise.

PR #700, opened 2026-08-21
fix(parser): harden Al Rajhi Saudi transaction parsing

8valet:
## Summary

Improve Al Rajhi transaction notification parsing for additional historical Saudi SAR/SR formats.

This change:

- adds support for additional SAR/SR amount layouts
- classifies refunds, reversals, and cashback corrections correctly
- rejects declined, failed, and non-transaction notifications
- improves merchant extraction from supported Arabic and English fields
- improves current-balance extraction without treating balances as transaction amounts
- adds synthetic positive and negative regression coverage

## Type of change

- [ ] feat: New feature
- [x] fix: Bug fix
- [ ] docs: Documentation update
- [ ] test: Add/update tests

## Screenshots/Recordings

Not applicable — parser-only change.

## How to test

Run the focused Al Rajhi parser tests:

```bash
./gradlew :parser-core:jvmTest --tests AlRajhiBankParserTest

### DIFF
--- a/parser-core/src/main/kotlin/com/pennywiseai/parser/core/bank/AlRajhiBankParser.kt
+++ b/parser-core/src/main/kotlin/com/pennywiseai/parser/core/bank/AlRajhiBankParser.kt
@@ -38,8 +38,14 @@ class AlRajhiBankParser : BankParser() {
 
     private companion object {
         // English PoS format patterns, compiled once.
-        val POS_DETECT = Regex("""Amount:\s*SR\s""", RegexOption.IGNORE_CASE)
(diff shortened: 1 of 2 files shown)


sarim2000:
Reviewed — merging. Nice coverage of the historical SAR/SR formats, refunds, cashback, and reversals.

One minor P2 for a future tightening (not blocking): \`isDeclinedOrFailedMessage()\` matches "declined/failed/rejected" against the whole message, so a *successful* refund/reversal/cashback SMS that happens to reference the original failed purchase would be suppressed. In practice Al Rajhi credit notifications don't recite the original decline, so this is an unlikely edge case — but if you want to harden it, gate the failure check so it only rejects when the message isn't also a refund/reversal/cashback (that wording already takes precedence elsewhere in the parser).

PR #708, opened 2026-08-22
fix(parser): improve Saudi parser reliability for SNB, STC, and D360

8valet:
## Summary

Improve Saudi transaction parsing reliability for SNB, STC/STCPAY, and D360 while keeping recognition conservative.

The changes cover:

- historical Arabic SAR/SR amount layouts for SNB
- refund, reversal, and cash-correction classification
- historical STC/STCPAY transaction formats
- outgoing SARIE transfer classification
- D360 SAR settlement extraction when foreign and SAR amounts coexist
- rejection of authentication, security, declined, failed, telecom, and non-transaction messages
- safe handling of valid credits that reference an earlier failed or declined transaction
- synthetic positive and negative regression coverage

## Type of change

- [ ] feat: New feature
- [x] fix: Bug fix
- [ ] docs: Documentation update
- [x] test: Add/update tests

## Screenshots/Recordings (optional)

Not applicable — parser-only change.

## How to test

1. Run the focused SNB, STC/STCPAY, and D360 parser tests.
2. Run `./gradlew :parser-core:test`.
3. Review the synthetic positive and negative regression cases.

Focused affected-parser tests: **40 passed**.

The full parser-core suite completed **1,572 tests** with two known documentation-sync failures:

- `BankSamplesDocTest`
- `SupportedBanksDocTest`

Both failures also reproduce on clean upstream `main`.

Repository-wide `./gradlew test` and `./gradlew lint` are currently blocked locally by the existing invalid Android SDK path in `local.properties`. The same SDK-resolution failure reproduces on clean upstream `main`.

## Breaking changes

- [x] No breaking changes
- [ ] Yes (describe):

## Related issues

None.

## Checklist

- [x] Tests added/updated (if applicable)
- [ ] Docs updated (if applicable)
- [ ] `./gradlew test` passes locally
- [ ] `./gradlew lint` passes locally

### DIFF
--- a/parser-core/src/main/kotlin/com/pennywiseai/parser/core/bank/D360BankParser.kt
+++ b/parser-core/src/main/kotlin/com/pennywiseai/parser/core/bank/D360BankParser.kt
@@ -27,34 +27,8 @@ class D360BankParser : BankParser() {
         return sender.uppercase().contains("D360")
     }
 
-    override fun extractAmount(message: String): BigDecimal? {
-        // Foreign purchase/withdrawal: "Amount: TRY 342.00 (SAR 27.51)" -> take the SAR conversion.
-        val convertedPattern = Regex(
-            """Amount\s*:\s*[A-Z]{3}\s*[0-9,]+(?:\.\d{1,2})?\s*\(\s*SAR\s*([0-9,]+(?:\.\d{1,2})?)\s*\)""",
-            RegexOption.IGNORE_CASE
-        )
-        convertedPattern.find(message)?.let { match ->
-            return parseAmount(match.groupValues[1])
-        }
-
-        // Local transaction: "Amount: SAR 250.00". Anchored to "Amount:" so the
-        // "Fee: SAR 0.00" line can never be mistaken for the transaction amount.
-        val localPattern = Regex(
-            """Amount\s*:\s*SAR\s*([0-9,]+(?:\.\d{1,2})?)""",
-            RegexOption.IGNORE_CASE
-        )
-        localPattern.find(message)?.let { match ->
-            return parseAmount(match.groupValues[1])
-        }
-
-        return null
-    }
-
-    private fun parseAmount(raw: String): BigDecimal? = try {
-        BigDecimal(raw.replace(",", ""))
-    } catch (e: NumberFormatException) {
-        null
-    }
+    override fun extractAmount(message: String): BigDecimal? =
+        FinancialMessageFields.sarSettlementOrAmount(message, listOf("Amount"))
 
     override fun extractTransactionType(message: String): TransactionType? {
         val lower = message.lowercase()
@@ -70,7 +44,7 @@ class D360BankParser : BankParser() {
     }
 
     override fun extractMerchant(message: String, sender: String): String? {
-        // Purchases and ATM withdrawals: "At: FEED ME" / "At: CITY,TR". Matched
+        // Purchases and ATM withdrawals: "At: SYNTHETIC MERCHANT" / "At: CITY,TR". Matched
         // case-insensitively (tolerates "AT:"), but the "at: <datetime>" line on
         // transfers is skipped by rejecting date-shaped captures.
         Regex("""At\s*:\s*([^\n]+)""", RegexOption.IGNORE_CASE).findAll(message).forEach { match ->
@@ -104,9 +78,9 @@ class D360BankParser : BankParser() {
 
     override fun isTransactionMessage(message: String): Boolean {
         val lower = message.lowercase()
-        if (lower.contains("otp") ||
-            lower.contains("verification code") ||
-            lower.contains("one time password")
+        if (SaudiTransactionMessageGuards.isDeclinedOrFailed(message) ||
+            SaudiTransactionMessageGuards.isPromotionalOrOperationalNotice(message) ||
+            FinancialMessageSafety.isSecurityCode(message)
         ) return false
 
         // Reject promotional messages that happen to mention transaction words
@@ -119,7 +93,7 @@ class D360BankParser : BankParser() {
         if (promoExact.any { lower.contains(it) } || PROMO_WORDS.containsMatchIn(lower)) return false
 
         val keywords = listOf(
-            "purchase", "withdrawal", "transfer", "incoming", "outgoing", "amount", "sar"
+            "purchase", "withdrawal", "transfer", "incoming", "outgoing", "account funding", "apple pay", "amount", "sar"
         )
         return keywords.any { lower.contains(it) }
     }
--- a/parser-core/src/main/kotlin/com/pennywiseai/parser/core/bank/FinancialMessageFields.kt
+++ b/parser-core/src/main/kotlin/com/pennywiseai/parser/core/bank/FinancialMessageFields.kt
@@ -1,0 +1,57 @@
+package com.pennywiseai.parser.core.bank
+
+import java.math.BigDecimal
+
+/** Conservative extraction for explicitly labelled financial fields. */
+internal object FinancialMessageFields {
+
+    enum class TransferDirection { INCOMING, OUTGOING }
+
+    fun sarAmount(message: String, labels: Collection<String>): BigDecimal? {
+        val value = labelledValue(message, labels) ?: return null
+        return sarAmountFromValue(value)
+    }
+
+    fun sarSettlementOrAmount(message: String, labels: Collection<String>): BigDecimal? {
+        val value = labelledValue(message, labels) ?: return null
+        SETTLEMENT_SAR_FIRST.find(value)?.let { return parse(it.groupValues[1]) }
+        SETTLEMENT_SAR_LAST.find(value)?.let { return parse(it.groupValues[1]) }
+        return sarAmountFromValue(value)
+    }
+
+    fun transferDirection(message: String): TransferDirection? {
+        val from = labelledValue(message, listOf("From")) != null
+        val to = labelledValue(message, listOf("To")) != null
+        return when {
+            from && !to -> TransferDirection.INCOMING
+            to && !from -> TransferDirection.OUTGOING
+            else -> null
+        }
+    }
+
+    private fun labelledValue(message: String, labels: Collection<String>): String? {
+        val alternatives = labels.joinToString("|") { Regex.escape(it).replace("\\ ", "\\s+") }
+        val pattern = Regex("""^(?:$alternatives)\s*:?\s*(.+?)\s*$""", RegexOption.IGNORE_CASE)
+        return message.replace("\r\n", "\n").replace('\r', '\n')
+            .lineSequence()
+            .map { it.trim() }
+            .firstNotNullOfOrNull { pattern.matchEntire(it)?.groupValues?.get(1)?.trim() }
+    }
+
+    private fun sarAmountFromValue(value: String): BigDecimal? {
+        SAR_FIRST.matchEntire(value)?.let { return parse(it.groupValues[1]) }
+        SAR_LAST.matchEntire(value)?.let { return parse(it.groupValues[1]) }
+        return null
+    }
+
+    private fun parse(raw: String): BigDecimal? = try {
+        BigDecimal(raw.replace(",", ""))
+    } catch (_: NumberFormatException) {
+        null
+    }
+
+    private val SAR_FIRST = Regex("""^(?:SAR|SR)\s*([0-9][0-9,]*(?:\.\d{1,2})?)$""", RegexOption.IGNORE_CASE)
+    private val SAR_LAST = Regex("""^([0-9][0-9,]*(?:\.\d{1,2})?)\s*(?:SAR|SR)$""", RegexOption.IGNORE_CASE)
+    private val SETTLEMENT_SAR_FIRST = Regex("""\(\s*(?:SAR|SR)\s*([0-9][0-9,]*(?:\.\d{1,2})?)\s*\)""", RegexOption.IGNORE_CASE)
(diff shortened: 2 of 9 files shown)


sarim2000:
Thanks for the broad Saudi coverage work. Holding merge on one real correctness issue Greptile flagged (agree it's blocking):

**Failure guard is bypassed by the historical-credit short-circuit.** In `SaudiTransactionMessageGuards.kt`, the historical-credit (refund/reversal → income) detection runs *before* the explicit failure/declined checks. So a **failed** refund or reversal that references a prior transaction gets classified as income and **persisted as a phantom credit** — worse than dropping a row, since it writes wrong data into the user's history.

**Fix:** run the failure/declined guard **first**, and only then the refund/reversal/credit classification (failure wording should always win). A regression case with a *failed* refund/reversal that mentions an earlier txn would pin it.

Rest of the PR (historical amount layouts, SARIE/correction/internal-transfer classes, the shared safety helpers) looks good and is well tested — just this ordering to sort before merge.

sarim2000:
Thanks for this! One Greptile finding to clear before merge (4/5): in `SaudiTransactionMessageGuards.kt` the historical-credit short-circuit runs before the explicit failure checks, so a *failed* refund/reversal that references a previous transaction can still be persisted as income. Could you move the failure checks ahead of that short-circuit? That should take it to 5/5.

sarim2000:
Thanks for this — the shared guards and the added coverage are a real improvement. One blocking issue before merge, which I've reproduced and have a fix for.

## Failed refunds are recorded as income

`isDeclinedOrFailed` returns early:

```kotlin
if (isHistoricalCreditCorrection(lower)) return false
```

That runs **before** any failure check, so any message containing a credit word *and* a historical word skips failure detection entirely. Run against `STCBankParser`:

| Message | Result |
|---|---|
| `Refund for your previous purchase failed. Amount SAR 250.00. Card *1234` | parsed, **type=INCOME, amount=250.00** |
| `Reversal of the original transaction was declined. SAR 100.00 Card *1234` | parsed, **type=INCOME, amount=100.00** |

Money that never arrived gets booked as income, and it inflates the account balance — worse than missing the transaction.

## Suggested fix

The intent behind the short circuit is right: a genuine refund often *describes* an earlier failed payment ("refund for your previously declined purchase"), and that mention shouldn't make the credit itself look failed. The problem is that it rescues the whole message instead of just that phrase.

Neutralise only the failure wording that is bound to the historical marker, then judge the message normally:

```kotlin
val scrubbed = redactHistoricalFailureMentions(message)
val lower = scrubbed.lowercase()
if (FinancialMessageSafety.hasExplicitFailure(scrubbed, ARABIC_FAILURES)) return true
// ...remaining checks against `scrubbed`
```

**Adjacency is the load-bearing part.** In "previously **declined** purchase" the failure describes the earlier purchase; in "previous purchase **failed**" a noun separates them and "failed" is this message's own verb. My first attempt allowed up to two words in between and it swallowed the second case — so the pattern requires the failure word directly against the marker, in either order, with Arabic equivalents.

Verified, all four cases now correct:

| | before | after |
|---|---|---|
| `...previous purchase failed` | INCOME 250.00 | **rejected** |
| `...original transaction was declined` | INCOME 100.00 | **rejected** |
| `...previously declined purchase. SAR 250.00 credited` | INCOME 250.00 | INCOME 250.00 |
| `Reversal of the original declined payment: SAR 100.00 credited` | INCOME 100.00 | INCOME 100.00 |

Full `parser-core` suite green — 126 test classes, no failures.

@8valet happy to push this to your branch if you'd rather (the PR allows maintainer edits), or you can take the approach and add it yourself with a couple of test cases for the failed-refund shapes. Your call.

sarim2000:
Pushed the fix to this branch as `1d7d2040` (the PR has maintainer edits enabled) — @8valet, amend or revert it freely if you'd rather take a different approach.

What landed:

- Only the failure wording *bound to* the historical marker is neutralised; the message is then judged normally, so a refund that itself failed is caught while a refund describing an earlier declined purchase still parses.
- Adjacency is what separates them, and it's the part that's easy to get wrong — my first attempt allowed a couple of words in between and it swallowed "previous **purchase** failed", which is exactly the bug.
- Added the two failed-refund shapes as rejection cases in `STCBankParserTest`, alongside your existing "Refund for previously declined purchase is accepted".

The tests are load-bearing rather than decorative: reverting the guard change makes **both fail** (16 tests, 2 failures), and with it they pass. Full `parser-core` suite green, no failures anywhere.

sarim2000:
Also right — pushed `8c963bb2`.

My adjacency rule was still too permissive. Both of these put the failure word directly against the historical marker:

- `Refund for your previously declined purchase` — "declined" describes the earlier **purchase**; the refund is real.
- `Your previously declined refund` — "declined" describes the **refund**; that credit never landed.

The noun after the failure word is what separates them, so redaction is now skipped when a credit noun (`refund`, `reversal`, `cashback`, `correction`) follows.

Probed against `STCBankParser` before and after:

| | before | after |
|---|---|---|
| `Refund for your previously declined purchase` | INCOME | INCOME |
| `Your previously declined refund` | **INCOME** | **rejected** |
| `Your previously failed reversal` | **INCOME** | **rejected** |
| `Refund declined` | rejected | rejected |

Both new shapes added as rejection cases — `STCBankParserTest` is at 18, full `parser-core` suite green with no failures.

sarim2000:
Pushed `968bc887` for this one.

`Refund was previously declined` put the credit noun *before* the failure phrase, so my lookahead never saw it and the redaction stripped the message's only failure status.

What decides the meaning is the **trailing qualifier**:

| Wording | What the failure describes | |
|---|---|---|
| `previously declined purchase` | the earlier purchase | refund is real |
| `previously declined refund` | this credit | never landed |
| `was previously declined` (nothing after) | this credit | never landed |

Redaction now requires a following non-credit word, **on the same line** — the next field's label (`Amount: ...`) was otherwise being read as the qualifier, which is exactly what let this shape through on my first attempt at it.

All seven known orderings now behave, with only the genuine refund accepted. `STCBankParserTest` at 19 cases, full `parser-core` suite green.

sarim2000:
Pushed `83a2ad3f`. This was my mistake compounding: I applied the qualifier rule to the forward word order and left the reverse branch redacting unconditionally, so `Refund declined earlier` and `Reversal failed previously` were still booked as income.

The rule is the same in both directions — **a failure is historical only when it describes some noun other than this credit.** What changes is which side that noun sits on:

- `previously declined **purchase**` — qualifier trails
- `**purchase** declined earlier` — qualifier leads

Both branches now require it, and reject when the neighbouring noun is the credit itself.

All eleven orderings I've tried, verified against `STCBankParser`:

| Message | |
|---|---|
| `Refund for your previously declined purchase` | accepted |
| `Refund for a purchase declined earlier` | accepted |
| `Purchase Reversal` | accepted |
| `Your previously declined refund` | rejected |
| `Your previously failed reversal` | rejected |
| `Refund was previously declined` | rejected |
| `Refund declined` / `Refund declined earlier` | rejected |
| `Reversal failed previously` | rejected |
| `Refund for your previous purchase failed` | rejected |
| `Reversal of the original transaction was declined` | rejected |

`STCBankParserTest` at 22 cases, full `parser-core` suite green.

Worth being straight about the trajectory: this guard is heuristic text-matching, and each round has found another word order rather than a flaw in the principle. If more shapes keep turning up, the safer design may be to invert the default — reject any credit message mentioning failure at all, and accept the "refund for a previously declined purchase" case only via an explicit allowlist. A missed refund is an annoyance; phantom income corrupts the balance. Happy to take that direction instead if you'd prefer it, @sarim2000 / @8valet.

sarim2000:
Pushed `0fee86b6` for the Arabic side. Two separate things were wrong:

**1. The Arabic branches had no qualifier requirement** — the same gap I'd just closed for English, left open in the other language. Arabic puts adjectives after the noun, so the qualifier always *leads* there:

- `عملية شراء سابقة مرفوضة` — a previous **purchase** that was rejected → the refund is real
- `استرجاع سابق مرفوض` — the **refund** itself → never landed

Both branches now require a leading non-credit word, anchored to a token boundary. That anchor matters: an unanchored `\S+` can start matching mid-word, so the credit-noun check was silently bypassed on my first attempt — it took a probe to notice.

**2. Nothing recognised a rejected credit even when the wording survived.** The Arabic failure phrases are keyed to debit nouns (`شراء`, `حوالة`, `سداد`…), so a bare `استرجاع … مرفوض` matched none of them and sailed through. A credit noun left next to a failure word after redaction now counts as a failure.

Your existing `استرجاع عملية شراء سابقة مرفوضة` case still parses as income, and all eleven English orderings are unchanged. Full `parser-core` suite green.

**One caveat I want to be explicit about, @8valet:** I don't read Arabic. The rule above is a *structural* inference from the one sample in your test file — noun then adjectives — and the vocabulary lists are extrapolated from the words already in this file. The English behaviour I can vouch for; the Arabic phrasing I'd like you to sanity-check, particularly whether `استرجاع سابق مرفوض` is natural wording for a rejected refund and whether there are common forms with the noun elsewhere in the sentence.

This is the fourth word-order variant found in this guard, which I think strengthens the point from my last comment: heuristic text-matching keeps yielding new shapes. Inverting the default — reject any credit message mentioning failure, allowlist the legitimate refund shape — would make the failure mode "missed refund" instead of "phantom income". Happy to do that instead if you'd both prefer.

sarim2000:
Pushed `999cf1a1`. This one changed the design rather than adding another pattern, because the previous shape showed the rule itself was wrong.

`Refund transaction declined earlier` passed: my rule only asked that the qualifying noun *not be a credit word*, and `transaction` satisfies that. But a generic noun doesn't say **whose** transaction — there it's the refund's own. Same for `العملية` in Arabic.

So the qualifier now has to name an actual kind of payment, from a closed list:

- `purchase | payment | transfer | withdrawal | charge | debit | pos`
- `شراء | حوالة | سداد | خصم | سحب | إيداع | مشتريات`

**Anything unrecognised is treated as the credit itself**, so the guard fails closed — an unfamiliar phrasing now costs a refund rather than inventing income. That's the substance of the inversion I raised in my last two comments, scoped so your accepted cases keep working rather than as a rewrite.

Seventeen orderings across both languages, all behaving:

| Accepted (4) | Rejected (13) |
|---|---|
| `Refund for your previously declined purchase` | `previously declined refund` · `previously failed reversal` |
| `Refund for a purchase declined earlier` | `Refund was previously declined` · `Refund declined` |
| `Purchase Reversal` | `Refund declined earlier` · `Reversal failed previously` |
| `استرجاع عملية شراء سابقة مرفوضة` | `previous purchase failed` · `original transaction was declined` |
| | `Refund transaction declined earlier` · `previously declined transaction` |
| | `استرجاع سابق مرفوض` · `استرجاع العملية سابقة مرفوضة` |

`STCBankParserTest` 23 cases, `SNBAlAhliBankParserTest` 11, full suite green.

@8valet the Arabic caveat from my previous comment still stands — the debit-noun list is extrapolated from words already in this file and I can't judge whether it's complete for real SNB/STC wording. If a common phrasing is missing, the cost is now a *rejected* refund rather than phantom income, which is the safer way round, but worth your eyes.

sarim2000:
Pushed `00945b09`. Good catch, and note the finding has **changed direction** — this is a dropped refund, not phantom income.

`حوالة` was in the qualifier list, `تحويل` wasn't, and both are ordinary words for a transfer. A refund referencing an earlier rejected `تحويل` therefore failed the qualifier check and was discarded as though the refund itself had failed.

That's the failure mode the closed list was chosen for: an incomplete list costs a real refund instead of inventing income. Still a genuine loss, so the word is now listed. Worth noting the parsers themselves only look for `حوالة` — the guard covers both on purpose, because it's reading whatever the bank actually wrote rather than the vocabulary we happen to detect on.

| | |
|---|---|
| `استرجاع تحويل سابق مرفوض` | accepted (was dropped) |
| `استرجاع حوالة سابقة مرفوضة` | accepted |
| `استرجاع عملية شراء سابقة مرفوضة` | accepted |
| `استرجاع سابق مرفوض` | rejected |
| `استرجاع العملية سابقة مرفوضة` | rejected |

`SNBAlAhliBankParserTest` at 12 cases, full suite green.

@8valet this is precisely the review I asked for in my earlier comment — the debit-noun lists are extrapolated and I can't tell whether they're complete for real SNB/STC wording. If other common forms are missing, the symptom is a silently missing refund. Your eyes on both lists would be worth more than another round of my guessing.

sarim2000:
Pushed `45d4cc67`, which **deletes** the historical-mention rescue rather than patching it again. Net −70 lines of regex.

Six attempts at deciding from the text whether a failure describes an earlier payment or this credit, six new word orders. Every structural signal has a counterexample:

| Signal | Beaten by |
|---|---|
| neighbouring noun isn't a credit word | `refund transaction declined earlier` — the transaction is the refund's |
| noun names a specific kind of payment | `refund payment declined earlier` — the payment *is* the refund |
| credit followed by a preposition | `reversal of the original transaction was declined` — the reversal failed |

"Refund of a payment that failed" and "refund that failed" are the same words reordered. The guard now picks a failure mode instead of pretending to resolve the ambiguity.

**This is a behaviour change, not a passing detail.** A genuine `Refund for a previously declined purchase` is now dropped, and five cases across STC, SNB and D360 that asserted acceptance now assert rejection — including ones you wrote, @8valet. I did it because the alternative was leaving a live phantom-income path (`Refund payment declined earlier` was still booking income), and a missing transaction is recoverable by hand while a corrupted balance isn't.

If you'd rather keep the rescue, revert this one commit — everything before it stands on its own. If dropped refunds turn out to matter in practice, the upgrade I'd suggest is an allowlist of *real* bank phrasings you've actually seen, not more grammar heuristics; that also sidesteps the Arabic problem I flagged earlier, where I was inventing wording I can't read.

Full `parser-core` suite green, no failures.

Issue #303, opened 2026-05-04
[FEATURE] Easy path to add/edit category selection from transaction alert notification.


tiniwings:
## 💡 Feature Description
Current: Click on notification -> Opens transaction in app > Click Edit at the top - > User Selects category.

Proposed: Provide Add/Edit Category button option in transaction notification - > Directly take user to transaction category section with edit mode -> User selects category.

Or **It would be great if you can directly provide a popup window to select category by clicking on notification.**

## 🎯 Problem it Solves
Allows user to add/edit category directly from notification with less number of steps

## 🔧 Proposed Solution
Small Popup window to select category from Notification action.

PR #710, opened 2026-08-23
feat(notification): inline tagging from the quick picker (#696)

### LINKED ISSUE #696: [FEATURE] custom tag
## 💡 Feature Description
 When payment is done, with the options already present like food and dining, income, other, there should also be an option to type out specific tag, so that it's differentiated, like even in food and dining option, there would be necessary spending, and spending on pleasure, so to differentiate anything like that, a feature of custom tag is needed, in the notification itself, instead of going inside the app and doing it.

sarim2000:
Closes #696.

## Request
Let users add a **custom tag** at capture time — "in the notification itself" — to differentiate spending within a category (e.g. Food & Dining → necessary vs pleasure), without opening the app.

## What already existed
Custom tags are a full in-app feature (create/apply/filter, `TagInputField`, `TagRepository`), and the txn-alert notification (#303) already has a **More…** action opening `QuickCategoryPickerSheet`. This PR brings tagging to that sheet — the notification's action bar is already full (2 category quick-picks + More…), so a new affordance there wasn't an option.

## Change (UI-only)
- `QuickCategoryPickerSheet` gains optional `tagSuggestions` / `initialTags` / `onTagsChanged`. When `onTagsChanged` is non-null, a **Tags** section (reusing `TagInputField`) renders above the category list. Each add/remove **commits immediately**, so tagging works even if the user dismisses without picking a category. In-app call sites omit these params → unchanged.
- `QuickCategoryPickerActivity` injects `TagRepository`, seeds the field from the txn's current tags + all tag names for autocomplete, and persists via `setTagsForTransaction`.

## Verification
`./init.sh app` green. UI change — worth an on-device look at the More… sheet from a transaction notification before merge.

### DIFF
--- a/app/src/main/java/com/pennywiseai/tracker/data/repository/TagRepository.kt
+++ b/app/src/main/java/com/pennywiseai/tracker/data/repository/TagRepository.kt
@@ -5,8 +5,13 @@ import com.pennywiseai.tracker.data.database.PennyWiseDatabase
 import com.pennywiseai.tracker.data.database.dao.TagDao
 import com.pennywiseai.tracker.data.database.entity.TagEntity
 import com.pennywiseai.tracker.data.database.entity.TransactionTagCrossRef
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.SupervisorJob
+import kotlinx.coroutines.channels.Channel
 import kotlinx.coroutines.flow.Flow
 import kotlinx.coroutines.flow.map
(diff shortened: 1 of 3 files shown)


sarim2000:
✅ On-device verified (PennyAI_Test emulator): injected a Kotak SMS → backgrounded → txn-alert notification's **More…** picker now shows the **Tags (Optional)** field above the category list; typed a tag and confirmed it persisted — DB shows `Bakery ↔ necessary` via `transaction_tag_cross_ref`.

PR #712, opened 2026-08-23
feat(home): option to count credit-card spend as expense (#705)

### LINKED ISSUE #705: [FEATURE] Option to count credit-card spend as Expense (not Credit)
From Play Store review feedback.

## Current behavior
Credit-card purchases are `TransactionType.CREDIT`. In `BalanceCalculator`, CREDIT increases the card's outstanding balance (correct for card accounting) and is kept **separate** from the Expenses tile / cash-flow. Some users instead want CC spend to count as regular spending/expense.

## Ask
A user-facing setting (e.g. Settings → "Count credit-card spend as expense") that folds CREDIT into expense totals on the spending surfaces (Home cash-flow, Analytics, budgets) for users who track it that way. Default off to preserve current behavior. Needs care so it doesn't double-count against the card's outstanding balance.

sarim2000:
Closes #705.

## Change
New **default-off** preference **"Count card spend as expense"** (Settings → Currency). When enabled, `TransactionType.CREDIT` is folded into the Home **"Spent this month"** expense total (`computeBreakdownByCurrency`), threaded through the breakdown flows via `combine` so toggling refreshes live.

- **Default off** → no behavior change. Because CREDIT is currently *excluded* from that expense figure, including it when on **can't double-count**.
- **Scope:** the fixed Home spending figure. The **Transactions** and **Analytics** screens keep their explicit per-type breakdowns on purpose — Analytics already lets you select the **Credit** type filter to view card spend, so it doesn't need the toggle. This avoids a half-applied semantic that would make surfaces disagree.

Files: preference (`UserPreferencesRepository`), `HomeViewModel` (filter + reactive wiring), `SettingsViewModel` + `SettingsScreen` (toggle).

## Verification
`./init.sh app` green. **On-device toggle check pending** — the shared emulator was being actively driven by another session at the time, so I didn't contend it; the change is a default-off filter tweak on the standard Settings-toggle pattern. Worth a quick on-device confirm (toggle on → "Spent this month" includes a current-cycle CC charge) before merge.

### DIFF
--- a/app/src/main/java/com/pennywiseai/tracker/data/preferences/UserPreferencesRepository.kt
+++ b/app/src/main/java/com/pennywiseai/tracker/data/preferences/UserPreferencesRepository.kt
@@ -61,6 +61,7 @@ open class UserPreferencesRepository @Inject constructor(
         val APP_FONT = stringPreferencesKey("app_font")
         val HAS_SKIPPED_SMS_PERMISSION = booleanPreferencesKey("has_skipped_sms_permission")
         val DEVELOPER_MODE_ENABLED = booleanPreferencesKey("developer_mode_enabled")
+        val COUNT_CREDIT_AS_EXPENSE = booleanPreferencesKey("count_credit_as_expense")
         val SYSTEM_PROMPT = stringPreferencesKey("system_prompt")
         val HAS_SHOWN_SCAN_TUTORIAL = booleanPreferencesKey("has_shown_scan_tutorial")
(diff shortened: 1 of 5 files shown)


sarim2000:
✅ On-device verified (PennyAI_Test emulator): with two current-cycle credit-card transactions (₹2,400 total), **"Spent this month" = ₹1,235 with the toggle OFF → ₹3,635 with it ON** (folds in the ₹2,400 CC spend), updating reactively on return to Home. The separate "Money in motion → Credit ₹2,400" breakdown still shows. Supersedes the earlier 'on-device pending' note.

PR #713, opened 2026-08-23
feat(parsers): add GTBank, Standard Chartered Nigeria, and VFD Microfinance Bank parsers

chikaezeani:
## Summary

Adds SMS parsers for three Nigerian banks — **GTBank**, **Standard Chartered Nigeria**, and **VFD Microfinance Bank** — taking Nigeria coverage from 5 to 8 banks (147 total, 23 countries).

Also fixes a bug where changing the base currency never reached the AI assistant.

All three parsers were validated on-device against a real Nigerian inbox, not just fixtures. An independent reconciliation of the parsed transactions reproduced the app's monthly totals exactly, with zero misparses, zero duplicates, and zero currency errors.

## Parsers

**GTBank** — line-based; direction is a `DR`/`CR` suffix on the `Amt` line rather than a leading `Debit`/`Credit` line, and labels are `Acct:`/`Bal:` (Access Bank uses `Acc:`/`Avail Bal:`).

```
Acct:******4321
Amt:NGN15,000.00 DR
Desc:OUTWARD TRANSFER TO OPAY - JANE DOE
Bal:NGN20,000.00
Date:2026-01-15 9:36AM
```

**Standard Chartered Nigeria** — single line, comma-separated, `Credit Alert!`/`Debit Alert!` prefix.

```
Credit Alert! Acct:xxxxxx1234, Amt:NGN1000.00, Desc:..., Date:2026-01-15, Bal:NGN1500000.00
```

This shares a sender ID (`StanChart`, `SCBANK`) with the existing India/Pakistan `StandardCharteredBankParser`. Rather than widen that parser, this one is registered ahead of it and gates strictly on the alert-prefix + `Amt:NGN` shape, which does not occur in the India/Pakistan formats. `BankParserFactory.parse`'s content-aware dispatch keeps both working — a test asserts a Nigerian message routes to Nigeria and an Indian UPI message still routes to India.

**VFD Microfinance Bank** — line-based, `N` prefix (not `NGN`), `DR`/`CR` suffix.

```
Acct: xxx901
Amt: N11,000.00 DR
Date: 15-JAN-2026 21:53:32
Chgs: N25.00 (COMM VAT)
Desc: Fruits/ To JANE DOE
Balance:N30,000.00
```

Note the `Chgs:` line carries its own amount. Every money pattern is anchored to its specific label, so the `N25.00` fee can never be read as the `N11,000.00` transaction — pinned by a dedicated test.

## Notes

- Descriptions capture with `[ \t]*` rather than `\s*`, so an empty `Desc:` cannot cross the newline and grab the following line as the merchant — same fix already applied to `AccessBankParser`.
- Two format variants found only in real messages are pinned by tests: VFD alerts that **omit the `Date:` line** entirely, and the real sender ID being **`VFD MFB`**, not a bare `VFD`.
- Adds `TestNigeriaEndToEnd` which goes through `BankParserFactory` the way the app does on an incoming SMS — covering registration, sender routing and dispatch, not just the regexes.
- Catalogue regenerated via `scripts/update-supported-banks.sh`.
- All sample data in tests and docs is fictional.

## The currency fix (second commit)

Changing the base currency had no effect on the AI assistant — it kept reporting the previous currency indefinitely.

The system prompt embeds pre-formatted amounts and is cached in DataStore. `LlmRepository.sendMessageStream` reuses the stored prompt whenever one exists, and nothing invalidated it on a currency change.

The only live refresh path is `OptimizedSmsReaderWorker`, gated on `stats.saved.get() > 0` — it fires only when a scan saves a *new* transaction. **For a user whose inbox is already fully scanned, every rescan saves zero rows, so the prompt is never rebuilt and the currency change can never take effect.** (`HomeViewModel.updateSystemPrompt()` would also do it, but it has no caller.)

The fix removes the cached prompt in the same atomic `edit` as the currency write. `applyMainAccountCurrency` gets the same treatment, but only when the value actually changes.

Reproduced and verified on-device before and after.

## Testing

- `./init.sh` — full gate passes
- 43 new tests across the three parsers plus the end-to-end suite; 1,593 total passing
- On-device against a real Nigerian inbox

## Not included

Two further observations from on-device testing that change existing behaviour and are better discussed separately:

1. **`baseCurrency` defaults to `"INR"`** rather than deriving from device locale, so a first-run user outside India starts on the wrong currency.
2. **The AI is asked about timeframes it has no data for.** The prompt carries monthly totals and 14 days of transactions but no weekly aggregate; asked "this week", the model answers with the monthly figure relabelled. A `Last 7 days` line plus a guideline not to reuse one period's figure under another label fixes it.

Happy to open these as issues, or fold either into this PR if you'd prefer.

_Correction: an earlier revision of this description claimed `TRANSFER` rows were being summed into income/expense totals. That was wrong — `toMonthlyBreakdown` filters on `transactionType` exactly, and `AiContextRepository` skips transfers explicitly, so neither path can include them. Apologies for the noise._


sarim2000:
Thanks for this! Just the one Greptile finding to clear before merge (4/5): the currency-change invalidation removes the DataStore prompt cache but not the system prompt already stored in Room chat history or the active LLM conversation, so an existing chat keeps answering in the old currency. Mind also invalidating those two in `UserPreferencesRepository.kt`? Should get it to 5/5.

chikaezeani:
Good catch — fixed in d7ddf59.

You're right that clearing DataStore wasn't enough. `ensureConversation()` rebuilds the LLM session from the `isSystemPrompt = true` row in Room, never from DataStore, so an existing chat never saw the new currency; and if the LiteRT-LM conversation was still live, even the Room copy wasn't re-read. Only a fresh chat picked up the change.

One deviation from your suggestion worth flagging: I put the invalidation in `LlmRepository` rather than `UserPreferencesRepository`. That class injects only a `Context`, so reaching Room and the LLM session from it would mean injecting `ChatDao` and `LlmService` into a preferences store — inverting the layering, and creating a Hilt cycle since `LlmRepository` already depends on `UserPreferencesRepository`. `LlmRepository` already owns all three pieces of state, so `invalidateSystemPrompt()` lives there and both currency entry points (Settings selector and main-account derivation) call it. Happy to move it if you'd rather keep the logic in one file.

Two details:

- `applyMainAccountCurrency()` now returns whether the currency actually changed. Without that the main-account path would clear chat history on every recompute, not just on a real change.
- Invalidation clears chat history wholesale rather than just the system row, since the prior answers quote the old currency too — leaving them would make the transcript self-contradictory. Say the word if you'd prefer to keep the visible messages and swap only the system prompt.

`./init.sh app` passes.

chikaezeani:
Both findings addressed in 1938c2d — but by changing the approach rather than patching the previous one.

Greptile was right that re-selecting the active currency deleted the whole chat. Stepping back, though, the destructive reset was the wrong remedy in the first place. The system prompt is *derived state* — a rendering of the user's finances in the chosen currency. A currency change invalidates the **rendering**, not the conversation. Deleting someone's chat to fix a formatting problem is too blunt.

So `invalidateSystemPrompt()` is now `refreshSystemPrompt()`, which re-renders in place:

- rebuilds the prompt against the new currency;
- swaps **only** the hidden `isSystemPrompt` row via a new `deleteSystemPromptMessages()` DAO query, which cannot touch visible messages;
- recreates the LiteRT-LM conversation replaying the same history — `systemInstruction` is fixed at construction, so a rebuild is the only way to change it, but the model keeps full context and continues under the new instruction.

The user-visible behaviour is now: change currency mid-conversation, come back to the chat, history intact, assistant has switched currency. No data loss either way.

On the two specific findings:

- **Unchanged currency deletes history** — `updateBaseCurrency()` now returns whether the currency actually moved, and the caller only re-renders on a real change. The comparison happens inside the DataStore `edit` transaction, so it can't race a concurrent write the way a separate read-then-write could.
- **Separate cancellable operations** — the re-render runs in `NonCancellable`. The durable currency write has already landed by that point, so leaving Settings mid-flight must not strand a stale prompt.

Added `LlmRepositoryPromptRefreshTest` covering history preservation, prompt replacement, ordering, conversation replay, the empty-chat no-op, and the unchanged-currency no-op.

One small production change worth flagging: `baseCurrency` and `updateSystemPrompt` are now `open`. `preferencesDataStore` caches one instance per process keyed by file name, so any test driving the real DataStore contends with every other test in the JVM — stubbing those two members avoids that. Happy to take a different approach if you'd rather not widen the class.

`./init.sh` passes (app + parser-core).

chikaezeani:
Closing this in favour of #733, which carries the three Nigerian parsers on their own.

Splitting them out was overdue — the parsers were never the contested part, and they shouldn't have been held up by the AI prompt work bundled alongside them.

On that AI work: I withdrew it rather than iterate again. The currency-invalidation fix was sound, but the concurrency fix I added on top of it did not hold up. Testing it properly showed it failed to close the race it targeted and introduced two new problems (a retired conversation could still be closed mid-stream because the in-flight counter is incremented at collection time, and a 5s drain timeout that would fire during normal generation rather than only on a wedged one). The tests I had written for it exercised a stand-in class rather than `LlmServiceImpl`, so they passed with the entire fix deleted — they were not catching any of it.

Better to leave a rare pre-existing crash alone than to ship a fix that makes it more likely.

Two things worth recording separately, both pre-existing and neither addressed here:

- `LlmRepository.updateSystemPrompt()` (called by `OptimizedSmsReaderWorker` after a scan saves transactions) updates the DataStore prompt and closes the conversation but never updates the `isSystemPrompt` row in Room — and `ensureConversation()` rebuilds from that row. So the chat can keep answering from a stale financial summary after a scan.
- `BackupImporter` clears `chat_messages` directly via the DAO, bypassing `LlmRepository`, so a live conversation keeps replaying history that no longer exists in the database.

Happy to open issues for either if they'd be useful. Thanks for the review on this one — the Greptile findings were right each time.

Issue #418, opened 2026-06-05
[BUG] Analytics screen ignores hidden accounts — hidden account still counted in reports


sarim2000:
## Bug
Hiding an account removes it from the **Home** balance/carousel, but the **Analytics** screen still includes that account's transactions in every total, category breakdown, merchant breakdown, and trend. A user who hides their business account still sees its spending in reports.

## Steps to reproduce
1. Manage Accounts → hide an account (e.g. a business account).
2. Home balances/carousel correctly exclude it.
3. Open Analytics → that account's transactions are still counted.

## Expected
Analytics should respect the `hidden_accounts` set the same way Home does.

## Code pointers
- `HomeViewModel.kt` (~lines 346–350): loads the `"hidden_accounts"` SharedPreferences set and applies `filterVisibleBalances(...)` to both the carousel and the totals.
- `AnalyticsViewModel.kt`: fetches `transactionRepository.getTransactionsBetweenDates(...)` with **no** reference to `hidden_accounts` — all accounts flow into the aggregates.
- The `"hidden_accounts"` key is used in HomeViewModel / RulesViewModel / ManageAccountsViewModel but **not** AnalyticsViewModel.

---
_From a single user's feedback (a finance professional, ~2 days in). Common theme across several of these: the Analytics screen has no notion of account/profile scoping._

Issue #414, opened 2026-06-03
[BUG] Unable to restore from a backup


shreyes-shalgar:
## 🐛 Bug Description
Backup not working

## 📱 Steps to Reproduce
1. Install th latest release from 2.15.54
2. Import data from the backup file created on the previous release
3. See error

## 📸 Screenshots

<img width="720" height="1600" alt="Image" src="https://github.com/user-attachments/assets/6d57504d-7e57-4f82-9b39-9b4a584e9fbf" />

PR #738, opened 2026-08-28
feat(categories): hide unused/default categories from the pickers (#736)

### LINKED ISSUE #736: [FEATURE] Hide/disable unused default categories from the picker
From user email feedback.

## Request
Let users remove or **disable** default categories they don't use, so the category picker only shows relevant ones.

## Current state (verified in code)
`CategoriesScreen` supports edit + swipe-to-delete, but **system (default) categories are blocked**:
- `CategoriesViewModel.deleteCategory` returns *"System categories cannot be deleted"* for `isSystem` categories.
- Editing is gated on `!category.isSystem`.
- There is **no hide/disable** field on `CategoryEntity`.

So default categories can neither be removed nor hidden today.

## Proposed approach
Add a **hide** toggle rather than allowing deletion of defaults (deleting a default is risky — existing transactions/rules may reference it). A hidden category:
- is filtered out of the category **pickers** (quick picker, add/edit, rules),
- stays in the DB so historical transactions keep their category and still show in analytics.

There's precedent: accounts already support hiding (see #418). Mirror that.

Notes for implementation:
- `CategoryEntity` is backup-serialized — any new column needs a Kotlin default (Hard Constraint #3) + a DB migration.
- Add an unhide affordance so it's reversible.

sarim2000:
From user email feedback. Closes #736.

## Problem
Users can't remove or disable **default** categories they don't use — `CategoriesViewModel` blocks editing/deleting system categories (*"System categories cannot be deleted"*), and there was no hide/disable at all. So unused defaults always clutter the picker.

## Approach — hide, not delete
Add a per-category **hide** toggle (works for system defaults too). Hiding is the safe move for defaults: the row stays in the DB, so existing transactions keep their category and still show in analytics — it's just filtered out of the **pickers**. There's precedent: accounts already support hiding (#418).

## Changes
- **`CategoryEntity`**: `is_hidden` column, defaulted (backup-safe per #414). DB **migration 59→60** (additive column) + exported `60.json` schema.
- **`CategoryDao` / `CategoryRepository`**: new visible-only picker queries (`getVisibleCategories`, `getVisible{Expense,Income}Categories`) + `setCategoryHidden`. **Existing queries untouched**, so the manager and analytics keep seeing every category.
- **Pickers** switched to the visible-only variants: the transaction-detail category picker and the notification `QuickCategoryPicker`.
- **Manage Categories**: a visibility toggle on every row (system included); hidden rows render dimmed so they can be unhidden.

## Verification
- `./init.sh app` green, incl. `BackupSchemaGuardTest` (backup default correct).
- On the emulator: the **v60 migration column is present** on the real DB, and hiding a default (Food & Dining) **removes it from the picker query** (`is_income=0 AND is_hidden=0`) while **keeping it in the manager query** (present, `is_hidden=1`, dimmed). The manager toggle calls exactly that `UPDATE`.

Note: full click-through of the manager UI was awkward on a freshly-cleared app (onboarding), so the picker-filter behavior was verified at the on-device data layer + compile; the toggle UI mirrors the existing edit/delete row affordances.

https://claude.ai/code/session_017JdBDF1AncNhc76a4EnTuZ

### DIFF
--- a/app/schemas/com.pennywiseai.tracker.data.database.PennyWiseDatabase/60.json
+++ b/app/schemas/com.pennywiseai.tracker.data.database.PennyWiseDatabase/60.json
@@ -1,0 +1,2054 @@
+{
+  "formatVersion": 1,
+  "database": {
+    "version": 60,
+    "identityHash": "7c585fcb1739e34c6c7d5953c295e54a",
+    "entities": [
+      {
+        "tableName": "transactions",
+        "createSql": "CREATE TABLE IF NOT EXISTS `${TABLE_NAME}` (`id` INTEGER PRIMARY KEY AUTOINCREMENT NOT NULL, `amount` TEXT NOT NULL, `merchant_name` TEXT NOT NULL, `category` TEXT NOT NULL, `transaction_type` TEXT NOT NULL, `date_time` TEXT NOT NULL, `description` TEXT, `sms_body` TEXT, `bank_name` TEXT, `sms_sender` TEXT, `account_number` TEXT, `balance_after` TEXT, `transaction_hash` TEXT NOT NULL DEFAULT '', `is_recurring` INTEGER NOT NULL, `is_deleted` INTEGER NOT NULL DEFAULT 0, `excluded_from_analytics` INTEGER NOT NULL DEFAULT 0, `created_at` TEXT NOT NULL, `updated_at` TEXT NOT NULL, `currency` TEXT NOT NULL DEFAULT 'INR', `from_account` TEXT, `to_account` TEXT, `reference` TEXT, `loan_id` INTEGER DEFAULT NULL, `loan_contribution` TEXT DEFAULT NULL, `receipt_path` TEXT DEFAULT NULL, `budget_category` TEXT DEFAULT NULL, `budget_impact_type` TEXT DEFAULT NULL, `group_id` INTEGER DEFAULT NULL, `profile_id` INTEGER DEFAULT NULL)",
+        "fields": [
+          {
+            "fieldPath": "id",
+            "columnName": "id",
+            "affinity": "INTEGER",
+            "notNull": true
+          },
+          {
+            "fieldPath": "amount",
+            "columnName": "amount",
+            "affinity": "TEXT",
+            "notNull": true
+          },
+          {
+            "fieldPath": "merchantName",
+            "columnName": "merchant_name",
+            "affinity": "TEXT",
+            "notNull": true
+          },
+          {
+            "fieldPath": "category",
+            "columnName": "category",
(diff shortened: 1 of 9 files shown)


sarim2000:
Fixed the toggle race in `4dc8d79e`, and rebased onto `main` (which has moved on — #744, #745, #746 landed).

The finding was right: the screen computed the next value from a captured `CategoryEntity`, and the in-flight guard cleared as soon as the DAO update returned. Between that write landing and the Room Flow reaching Compose, the captured row is stale, so a second tap in that window asked for the same value again — a quick hide-then-show left the category hidden.

Rather than widen the guard, I removed the stale value entirely. `toggleCategoryHidden(id)` takes only the id and flips the flag in SQL:

```sql
UPDATE categories SET is_hidden = NOT is_hidden WHERE id = :categoryId
```

There's no value for the UI to get wrong, and the snackbar takes its wording from the row *after* the write. The in-flight guard stays, now purely to stop a burst of taps queueing overlapping writes.

Verified on the emulator against the database rather than the UI:

| | snackbar | `is_hidden` |
|---|---|---|
| Two taps ~120ms apart | "Entertainment shown" | **0** — reversed, as it should |
| One tap | "Entertainment hidden" | **1** |

sarim2000:
Right again, and this one I should have caught myself — fixed in `2d6cc637`.

The SQL inversion removed the stale *value*, but the in-flight guard still returned early while a write was running, so a second tap in that window was silently discarded. Two quick taps landed as one flip and the row ended up in the opposite state from the user's last tap.

My earlier emulator check passed only by luck: the two taps were ~120ms apart and the first write had already finished, so nothing was dropped.

Each category now has its own `Mutex` — taps on the same row queue behind each other and every one flips the row exactly once, instead of being thrown away.

Re-verified with both taps issued in a **single** adb round trip, so they genuinely overlap the write:

| Taps | `is_hidden` | |
|---|---|---|
| 2 | **0** | back to visible |
| 3 | **1** | hidden — correct parity |

sarim2000:
Fixed in the latest push — and the fix was cheaper than I expected, because the codebase already had the right tool.

The queued flip ran on `viewModelScope`, so tapping twice and immediately leaving the screen cleared the ViewModel and cancelled the second write, persisting the opposite of the user's last tap. It now runs on the existing `@ApplicationScope`, which `ApplicationModule` provides for precisely this — "fire-and-forget work that must outlive its launching component".

Verified with the taps and the back press issued in a **single** adb round trip, so the navigation really does race the writes:

| Action | `is_hidden` | |
|---|---|---|
| baseline | 1 | |
| 2 taps + navigate away | **1** | both flips landed |
| 1 tap + navigate away | **0** | write survived the screen |

PR #745, opened 2026-09-01
feat(settings): add a 'Delete All Transactions' action (#737)

### LINKED ISSUE #737: [FEATURE] Bulk 'delete all transactions' in Settings
From user email feedback.

## Request
A one-tap way to clear all recorded transaction history, without clearing the whole app's data or deleting entries one by one.

## Current state (verified in code)
`TransactionRepository.deleteAllTransactions()` / `TransactionDao.deleteAllTransactions()` exist but are **internal only** (used by backup import). There is **no user-facing action** in Settings — users currently must clear app data or delete transactions individually.

## Proposed approach
Add a **Settings → Data** action "Delete all transactions" behind a strong confirmation dialog (destructive, irreversible; suggest a "type to confirm" or double-confirm, and point users at Backup first).

Scope decisions to make:
- Transactions only, or also derived data (account balances, budgets snapshots)?
- Preserve user-curated links (loans/groups) like the full-resync path (`deleteUncuratedTransactions`), or a true wipe of everything?
- Offer it per-account too, or all-only? (Start with all-only.)

Recommend the smallest useful version first: wipe all transactions + their derived balances, with a clear confirmation.

sarim2000:
Closes #737.

## What

A one-tap way to clear transaction history, in **Settings → Data Management → Delete All Transactions**.

`TransactionRepository.deleteAllTransactions()` already existed but was only reachable from backup import, so users had to clear the app's data or delete rows one at a time.

## Scope decisions

The issue left two open questions:

- **Transactions only.** Accounts, budgets, loans, categories and rules are kept — the ask was to clear *history*, not to factory-reset. The dialog says so.
- **No loan/group preservation.** Unlike the full-resync path (`deleteUncuratedTransactions`), "delete all" means all. The cascading foreign keys on splits, tags and rule applications clear what hangs off the deleted rows; loans and groups survive as empty containers.

Widgets are refreshed afterwards, and the recent-transactions widget cache is cleared.

## The confirmation

Irreversible and unbatched, so a single tap isn't enough:

- names the exact row count, read from a new `countAllTransactions()` (soft-deleted rows included, so it matches what the delete actually removes)
- requires the word `DELETE`, case-sensitive
- points at Export Data first

## Verification

`./init.sh app` green.

On the emulator with 4 transactions:

1. The dialog reads "This permanently deletes all 4 transactions…" — the real count.
2. Typing `delete` (lowercase) leaves **Delete All** disabled; `DELETE` enables it.
3. Confirming reports "4 transactions deleted."; the Transactions tab under **All Time** is empty.
4. Home still shows the bank account and its balance — accounts survived.

### DIFF
--- a/app/src/main/java/com/pennywiseai/tracker/data/database/dao/AccountBalanceDao.kt
+++ b/app/src/main/java/com/pennywiseai/tracker/data/database/dao/AccountBalanceDao.kt
@@ -57,6 +57,39 @@ interface AccountBalanceDao {
         ORDER BY ab1.balance DESC
     """)
     fun getAllLatestBalances(): Flow<List<AccountBalanceEntity>>
+
+    /**
+     * One-shot form of [getAllLatestBalances], for callers that need the account
+     * list *inside* a database transaction — collecting a Flow there would
+     * observe the very writes the transaction is making.
+     */
+    @Query("""
+        SELECT ab1.* FROM account_balances ab1
+        INNER JOIN (
+            SELECT bank_name, account_last4, MAX(timestamp) as max_timestamp
+            FROM account_balances
+            GROUP BY bank_name, account_last4
+        ) ab2
+        ON ab1.bank_name = ab2.bank_name
+        AND ab1.account_last4 = ab2.account_last4
+        AND ab1.timestamp = ab2.max_timestamp
+        ORDER BY ab1.balance DESC
+    """)
+    suspend fun getAllLatestBalancesOnce(): List<AccountBalanceEntity>
+
+    /**
+     * Removes balance rows the app derived from a transaction — the signed delta
+     * snapshots written by `insertBalanceDelta`, which carry a `transaction_id`
+     * and no `sms_source`.
+     *
+     * Deliberately spares anything with an `sms_source`: a bank-reported balance
+     * is the bank's own word even though it arrived alongside a transaction, so a
+     * wipe of local history must not rewrite it.
+     *
+     * @return the number of rows removed.
+     */
+    @Query("DELETE FROM account_balances WHERE transaction_id IS NOT NULL AND sms_source IS NULL")
+    suspend fun deleteTransactionDerivedBalances(): Int
     
     @Query("SELECT * FROM account_balances ORDER BY timestamp DESC")
     fun getAllBalances(): Flow<List<AccountBalanceEntity>>
--- a/app/src/main/java/com/pennywiseai/tracker/data/database/dao/TransactionDao.kt
+++ b/app/src/main/java/com/pennywiseai/tracker/data/database/dao/TransactionDao.kt
@@ -125,6 +125,14 @@ interface TransactionDao {
     suspend fun deleteAllTransactions()
 
     /**
+     * Every row in the table, soft-deleted ones included, so the "delete all"
+     * confirmation reports the same number of rows [deleteAllTransactions]
+     * will actually remove.
+     */
+    @Query("SELECT COUNT(*) FROM transactions")
+    suspend fun countAllTransactions(): Int
+
+    /**
      * Deletes only transactions that don't carry user-curated annotations
      * (no loan link, no group membership). Used by the full-rescan path so
      * rebuilding the SMS-derived view doesn't blow away loans + their
--- a/app/src/main/java/com/pennywiseai/tracker/data/database/dao/TransactionDao.kt
+++ b/app/src/main/java/com/pennywiseai/tracker/data/database/dao/TransactionDao.kt
@@ -135,6 +135,14 @@ interface TransactionDao {
     @Query("SELECT COUNT(*) FROM transactions")
     fun observeAllTransactionCount(): Flow<Int>
 
+    /**
+     * One-shot row count, read *inside* the delete-all transaction to check that
+     * the table still holds exactly what the user confirmed before anything is
+     * removed.
+     */
+    @Query("SELECT COUNT(*) FROM transactions")
+    suspend fun countAllTransactions(): Int
+
 
     /**
      * Deletes only transactions that don't carry user-curated annotations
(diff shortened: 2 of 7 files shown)


sarim2000:
Both findings were real — pushed `ff52809d`.

**Account balances.** Confirmed: a normal delete inserts a reverting balance delta via `applyDeleteBalanceShift`, and the raw `DELETE FROM transactions` skipped it entirely.

I first routed delete-all through `DeleteTransactionUseCase(list, hardDelete = true)` — literally consistent with per-row delete — and measured it on the emulator: deleting 34 transactions moved the displayed balance from **₹50,046.07\*** to **₹3,02,789.07\***, because every expense got unwound at once. An inflated balance right after "clear my history" reads as a bug, so I didn't keep that.

What shipped instead splits by account kind:

- **SMS-tracked accounts keep their last reported figure.** That number is the bank's own word, not something derived from local history, so clearing the history shouldn't rewrite it.
- **Manual / cash accounts are recomputed** back to their opening balance. Their balance *is defined* as opening + Σ(transactions), so leaving it was the actual bug. `ensureManualOpening` runs **before** the delete — the opening is back-solved from the transaction sum that's about to disappear — and `recomputeManualBalance` runs after.

Verified: 3 transactions deleted, all 10 accounts kept, aggregate balance held at ₹3,02,789.07\* instead of shifting again.

**Stale count.** `deleteAllTransactions()` now returns its affected-row count and the result message uses that, so the dialog's number is only ever a preview and a row written by a background scan mid-dialog is both deleted and reported. There's no read-then-delete window left.

sarim2000:
Round 2 — the atomicity finding is right, fixed in `1fe36b98`.

The three steps (ensure-opening → DELETE → recompute) were three separate database transactions, so a crash in between could leave the transactions gone but a manual account still displaying a total for them. Extracted `DeleteAllTransactionsUseCase`, which wraps all three in one `withTransaction`, mirroring how `DeleteTransactionUseCase` already does it. The ViewModel keeps only UI concerns.

Re-verified on the emulator: 3 transactions deleted, 10 accounts kept, balance held at ₹3,02,789.07* — same as before the refactor, now atomic.

On the second point — a background SMS scan inserting a row after the DELETE — I'm not treating that as a defect. The scan legitimately finds a new message after the wipe; the user's history *was* deleted, and the reported count is the DELETE's own affected-row count, so it's accurate about what it removed. Blocking ingestion for the duration would be a bigger hammer than the problem. The single transaction narrows the window to the delete itself anyway.

sarim2000:
Round 3 — I'm not making further changes for these. Taking them in turn:

**Manual credit cards aren't recomputed.** True, and it's pre-existing behavior this PR doesn't touch. `isManualAccount` excludes credit cards on `main` today, with a stated reason:

> Credit cards are excluded — their balance is outstanding owed (spending increases it), which the income-positive recompute would get backwards.

`AccountBalanceRepository` isn't in this PR's diff (5 files: `TransactionDao`, `TransactionRepository`, `DeleteAllTransactionsUseCase`, `SettingsScreen`, `SettingsViewModel`). The same gap applies to deleting a single transaction, since `applyDeleteBalanceShift` routes through the same helper. Making delete-all the one place that recomputes credit cards would mean writing sign-aware logic the codebase deliberately avoided, in a path with no test coverage for it — worth doing on its own if it's a real problem, not smuggled into this PR.

**"The confirmation can authorize more deletion than it displays."** The action is "Delete All Transactions" and the dialog leads with "This permanently deletes **all** N transactions". The number is context for the blast radius, not a contract that exactly N rows die; consent is to *all*, and the result reports the true count. Freezing the count would mean deleting by id list, which reintroduces the exact non-atomicity round 2 asked me to remove.

**Concurrent ingestion after the delete.** Same point as round 2. A scan finding a new SMS after the wipe isn't the wipe failing — the history was deleted, and the count reported is the DELETE's own affected rows.

Rounds 1 and 2 caught two genuine defects (the unsettled balance ledger, the non-atomic multi-step mutation) and both are fixed and re-verified on device. I think this is done; happy to revisit if @sarim2000 disagrees on any of the three.

sarim2000:
Round 3 — you were right about the credit cards, and my first attempt at it was wrong. Fixed properly in `682163f0`; I also took the other two points.

**1. Manual credit cards.** Confirmed against the database, not just the UI. A card seeded at ₹5,000 with a ₹1,000 spend showed ₹6,000, and delete-all left it at ₹6,000 — exactly as you said.

My first fix tried to settle it by writing `current = opening`, which did nothing: the card had no OPENING row at all, and my "is this locally maintained?" test looked at the *latest* balance row, which by then was a `TRANSACTION` delta. Both helpers skipped it.

The gap is orphaned rows, not a missing recompute. Every transaction landing on an SMS-tracked account *or* a manual credit card leaves a derived delta snapshot (`source_type` `TRANSACTION`, `sms_source` null); once the transactions are wiped, those rows are orphans still holding the displayed balance. Delete-all now drops them, which settles credit cards with no sign-aware arithmetic at all. Rows carrying an `sms_source` are spared — an SMS-parsed balance is the bank's own word even though it arrived alongside a transaction.

Verified end to end:

| | before | after |
|---|---|---|
| Manual credit card ••9911 | ₹6,000, 2 orphaned delta rows | **₹5,000**, orphans gone |
| Bank-reported rows (`sms_source` set) | 22 | **22** — untouched |
| Transactions | 22 | 0 |

**2. Preview count.** Now observed rather than snapshotted — `observeAllTransactionCount()` — so a row a background scan writes while the dialog is open moves the number the user is consenting to.

**3. Concurrent ingestion.** The report is now accurate instead of silent: after the delete commits, a fresh count is read and the message appends "N new transactions arrived while it ran." I still don't think a scan finding new SMS after the wipe is the wipe failing, but reporting a clean sweep while rows exist was fair to call out.

sarim2000:
Round 4 — took this one too, in `d9b72d60`.

The delete removed whatever the table held, which need not be what the confirmation displayed. Two changes so it removes exactly what was authorised or nothing at all:

- The dialog's count is **observed** rather than snapshotted, so the number tracks rows arriving underneath the user.
- `invoke()` takes the confirmed count and re-checks it **inside the same transaction** as the delete. A mismatch returns `CountChanged`, deletes nothing, and tells the user the new figure to re-confirm.

Verified on device: with the dialog open showing "all 2 transactions", I injected a bank SMS; the dialog moved to **"all 3 transactions"** on its own, and confirming reported "3 transactions deleted" with the database left at 0 transactions, 0 orphaned delta rows, and the manual card still settled at ₹5,000.

PR #746, opened 2026-09-01
feat(rules): export and import smart rules as JSON (#741)

### LINKED ISSUE #741: [FEATURE] Export or import smart rules
## 💡 Feature Description 
Currently smart rules are something that we have to create for ourself and cannot be shared between peers in the community

## 🎯 Problem it Solves
The tedious job of configuring the smart rules.
This will allow community to store/share various templates for smart rules. 

## 🔧 Proposed Solution
Export the smart rules into json (or any other file format suitable) 

## 🎨 UI/UX Suggestions (Optional)
In smart rules ability of selecting multiples rules and export button

sarim2000:
Closes #741.

## What

**Export rules** / **Import rules** in the Smart Rules overflow menu, writing a small versioned JSON file through SAF. Nothing leaves the device unless the user shares the file themselves.

```json
{
  "version": 1,
  "rules": [
    {
      "name": "EmulatorRoundTrip",
      "description": null,
      "priority": 100,
      "conditions": [
        { "field": "AMOUNT", "operator": "LESS_THAN", "value": "200", "logicalOperator": "AND" }
      ],
      "actions": [
        { "field": "CATEGORY", "actionType": "SET", "value": "Food & Dining" }
      ],
      "isActive": true
    }
  ]
}
```

## Decisions

- **The portable form is narrower than `TransactionRule`.** Ids, timestamps and the system-template flag are facts about one install, so they're dropped on export and regenerated on import.
- **Built-in templates aren't exported.** Every install already ships them; exporting would only create duplicates on the other side.
- **Imported rules get fresh ids**, so a file can be applied more than once. A rule whose *name* already exists is skipped rather than duplicated.
- **The free-tier cap still applies.** Import can't push a free user past `FreeTierLimits.MAX_RULES` — otherwise it'd be a way around the paywall the create screen enforces. The result message says how many were held back.
- **A bad file is refused outright, not half-applied.** Unknown keys are ignored (forward compatible), but a newer `version`, a non-rule-set file, an oversized file, or one containing *any* incomplete entry is rejected and nothing is imported ("1 rule in this file is incomplete, so nothing was imported."). Importing most of a shared rule set would leave someone with a half-applied config they can't easily tell from the real thing. Repeats of a name *within* one file are the one thing collapsed rather than refused — the file is unambiguous there, first one wins — and the count is reported.
- **Not included: multi-select export.** The issue suggested selecting rules to export; that means a selection mode across the whole screen. The complaint is "rule sets can't be shared", which export-all + import solves — happy to add selection later if people ask.

The Reset action moved into the same overflow menu rather than adding a third top-bar icon.

## Verification

`./init.sh app` green, plus 7 new tests in `RuleSharingCodecTest` (round trip, fresh ids, templates excluded, bad file, newer version, invalid rule, unknown keys).

End-to-end on the emulator:

1. With only built-in templates → "You don't have any custom rules to export yet." (no picker, no empty file).
2. Created a custom rule → export writes the JSON above.
3. Importing that same file again → "Imported 0 rules. 1 already existed."
4. Deleted the rule, imported the file → "Imported 1 rule." and it's back on the list.

### DIFF
--- a/app/src/main/java/com/pennywiseai/tracker/data/repository/RuleRepositoryImpl.kt
+++ b/app/src/main/java/com/pennywiseai/tracker/data/repository/RuleRepositoryImpl.kt
@@ -61,6 +61,10 @@ class RuleRepositoryImpl @Inject constructor(
         ruleDao.insertRule(ruleToEntity(rule))
     }
 
+    override suspend fun insertRules(rules: List<TransactionRule>) {
+        ruleDao.insertRules(rules.map { ruleToEntity(it) })
+    }
+
     override suspend fun updateRule(rule: TransactionRule) {
         ruleDao.updateRule(ruleToEntity(rule))
     }
--- a/app/src/main/java/com/pennywiseai/tracker/data/rules/RuleSharing.kt
+++ b/app/src/main/java/com/pennywiseai/tracker/data/rules/RuleSharing.kt
@@ -1,6 +1,7 @@
 package com.pennywiseai.tracker.data.rules
 
 import com.pennywiseai.tracker.domain.model.rule.RuleAction
+import com.pennywiseai.tracker.domain.model.rule.isExecutable
 import com.pennywiseai.tracker.domain.model.rule.RuleCondition
 import com.pennywiseai.tracker.domain.model.rule.supportedOperators
 import com.pennywiseai.tracker.domain.model.rule.TransactionRule
@@ -130,12 +131,13 @@ object RuleSharingCodec {
      * GREATER_THAN, would sail through the shape check and be persisted as a
      * rule that can never match. So each part is validated too, and each
      * condition's operator is checked against [supportedOperators] — the same
-     * list the rule editor builds its picker from.
+     * list the rule editor builds its picker from — and each action against
+     * [isExecutable], so a rule the engine would silently no-op never lands.
      */
     private fun TransactionRule.isFullyValid(): Boolean =
         validate() &&
             conditions.all { it.validate() && it.operator in supportedOperators(it.field) } &&
-            actions.all { it.validate() }
+            actions.all { it.validate() && it.isExecutable() }
 
     fun decode(text: String): DecodedRuleSet {
         require(text.length <= MAX_FILE_BYTES) {
--- a/app/src/main/java/com/pennywiseai/tracker/data/rules/RuleSharing.kt
+++ b/app/src/main/java/com/pennywiseai/tracker/data/rules/RuleSharing.kt
@@ -59,6 +59,15 @@ object RuleSharingCodec {
     const val FILE_EXTENSION = "json"
 
     /**
+     * Largest file [decode] will look at. An exported rule set runs to a few
+     * hundred bytes per rule, so this holds thousands of them — while stopping
+     * a mis-picked video from being pulled into memory whole by a picker that
+     * has to accept every MIME type (see RulesScreen). A heap blow-up here would be an
+     * OutOfMemoryError, which the import's `catch (Exception)` would not catch.
+     */
+    const val MAX_FILE_BYTES = 1L * 1024 * 1024
+
+    /**
      * The exportable subset of [rules]: the user's own rules. Built-in
      * templates are excluded — every install already ships them, so sharing
      * them would only create duplicates on the other side.
@@ -82,14 +91,34 @@ object RuleSharingCodec {
     )
 
     /**
-     * Parses [text] into the rules it holds, keeping only entries that would
-     * pass [TransactionRule.validate] — a hand-edited or truncated file
-     * shouldn't be able to insert a rule the create screen would reject.
-     *
-     * @throws IllegalArgumentException if the file isn't a rule set, is empty,
-     *   or was written by a newer format version.
+     * What a file yielded: the rules worth importing, plus what was thrown away
+     * getting there. The counts are reported back to the user — a file that
+     * silently loses half its rules is worse than one that says so.
      */
-    fun decode(text: String): List<TransactionRule> {
+    data class DecodedRuleSet(
+        val rules: List<TransactionRule>,
+        /** Entries that wouldn't pass [TransactionRule.validate]. */
+        val invalid: Int,
+        /** Entries dropped because an earlier entry in the same file had that name. */
+        val duplicatedInFile: Int
+    )
+
+    /**
+     * Parses [text] into the rules it holds.
+     *
+     * Entries that wouldn't pass [TransactionRule.validate] are dropped rather
+     * than inserted — the create screen would reject them too — and so are
+     * repeats of a name that already appeared earlier in the same file, which
+     * would otherwise land as two rules the user can't tell apart. Both are
+     * counted so the caller can say what happened.
+     *
+     * @throws IllegalArgumentException if the file isn't a rule set, holds no
+     *   usable rule, or was written by a newer format version.
+     */
+    fun decode(text: String): DecodedRuleSet {
+        require(text.length <= MAX_FILE_BYTES) {
+            "That file is too large to be a rules file."
+        }
         val parsed = try {
             json.decodeFromString<SharedRuleSet>(text)
         } catch (e: Exception) {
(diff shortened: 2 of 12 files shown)


sarim2000:
All three findings were real — pushed `a81b3663`.

**Partial imports of mixed files.** `decode` filtered invalid entries and only threw when *none* survived, so a half-bad file imported quietly. It now returns a `DecodedRuleSet` carrying the rules plus the counts of what was dropped, and the result dialog says so: "Imported 1 rule. 2 entries in the file were unusable and were skipped."

I kept it lenient rather than rejecting atomically — one malformed entry shouldn't cost someone the other 20 rules — but it's no longer silent.

**Intra-file duplicate names.** Right: I deduped against *existing* rule names but not within the file. Later repeats of a name are now dropped (case-insensitive) and counted the same way.

**Unbounded read.** Fair, and the `catch (Exception)` genuinely wouldn't have caught the resulting `OutOfMemoryError`. The size is now checked up-front via `OpenableColumns.SIZE`, and the read is capped at 1 MB regardless for providers that don't report one — that's still thousands of rules at a few hundred bytes each.

Verified on the emulator:

- A file with 1 good rule, 1 with no actions, and a case-different repeat → "Imported 1 rule. 2 entries in the file were unusable and were skipped."; only `GreptileCheck` appears on the list.
- A 4 MB file → "That file is too large to be a rules file.", no OOM in logcat.

Codec tests are up from 7 to 10, covering mixed validity, in-file duplicates, and the size cap.

sarim2000:
Round 2 — this one I'd push back on, and the confusion is my fault.

The finding says mixed files are "partially imported instead of being rejected as the feature contract promises". There's no such promise in the code: `decode`'s KDoc documents dropping invalid entries and counting them, and `@throws` lists only three refusal cases (not a rule set / no usable rule / newer version). What *did* promise refusal was a line in this PR's description — "A malformed file is refused, not half-read" — written before I made the behavior lenient in `a81b3663`. I've corrected the description.

The behavior is deliberate: one malformed entry shouldn't cost someone the other twenty rules in a shared file. What the previous round correctly flagged was that the dropping was **silent**, and that's fixed — the result now reads "Imported 1 rule. 2 entries in the file were unusable and were skipped."

Refusing outright is still reserved for files that are wholly unusable: not a rule set, no usable rule at all, oversized, or a newer format version.

Happy to switch to all-or-nothing if you'd rather, @sarim2000 — it's a one-line change in `decode` — but I don't think it's the better default for a format whose whole point is being passed between people.

sarim2000:
Changed my mind on this one — `539f9225` makes it all-or-nothing.

An entry that wouldn't pass `validate()` now fails the whole file: *"1 rule in this file is incomplete, so nothing was imported."* You're right that a half-applied rule set is the worse failure mode; the person who exported it can send a good file.

The one thing still collapsed rather than refused is a name repeated **within** one file — the file is unambiguous about what that rule should be (first one wins), and the count is still reported.

Verified on the emulator: deleted `GreptileCheck`, then imported a file containing it alongside one incomplete entry. The file was refused and `GreptileCheck` did **not** come back — nothing partially applied. A valid file still imports normally.

PR description updated; codec tests still 10/10, with the mixed-file case now asserting the valid rule does *not* slip through.

sarim2000:
Round 3 — real gap, fixed in `8db6e950`.

`TransactionRule.validate()` only checks that the name, conditions and actions are *present* — it never asks the conditions and actions whether they're well-formed. Fine for the create screen, which builds them from pickers, but an imported file is arbitrary text: a condition like `AMOUNT / LESS_THAN / "abc"` passed the shape check and would have been persisted as a rule that can never match.

`decode()` now requires every condition and action to pass its own `validate()`, so such a file is refused like any other unusable one. Scoped to the importer rather than widening `TransactionRule.validate()`, which the create screen shares.

Two tests added for it (malformed AMOUNT comparison, blank SET value); codec tests now 12/12.

sarim2000:
Round 4 — both taken, in `38032cae`.

**Field/operator pairing.** `RuleCondition.validate()` checks the value, not whether the operator means anything for the field, so `MERCHANT / GREATER_THAN` imported cleanly and could never fire.

The rule editor already knew the answer — it was a private list in `CreateRuleScreen`. That list moves to the domain as `supportedOperators(field)` and is now the single source of truth for both the picker and the importer, so they can't drift. Only the wording stays in the screen, since one operator reads differently per field (`<` for an amount, `before` for a time). Verified on device that the picker is unchanged for all three label styles: Amount `< > =`, Merchant `contains / equals / starts with`, Time of Day `before / after / at or after / at or before / exactly at`.

**Partial persistence.** Imported rules were inserted one at a time. They now go through `RuleDao.insertRules`, which Room runs as a single transaction.

**Not done: action/field pairing.** Unlike operators, the app has no existing authority for which actions suit which field — the editor doesn't constrain it either. Inventing a matrix here would be new semantics rather than enforcing the editor's own rules, so I've left it. Happy to add one if @sarim2000 wants to define it.

Codec tests now 14/14, including the nonsensical-operator case.

sarim2000:
Round 5 — both specifics were right, fixed in `9cb29e14`.

Two values `RuleCondition.validate()` waved through, each of which `RuleEngine` can never match:

- **`AMOUNT` with `EQUALS`** skipped the numeric check entirely — only the four *ordering* operators were checked, everything else fell through to `true`. So `AMOUNT = abc` validated, and `=` is an operator the editor offers for amounts, so it was reachable. Every amount comparison is numeric now, with `IN`/`NOT_IN` checked as a comma-separated list.
- **`TRANSACTION_DATE` had no case at all.** The engine compares it against `LocalDate.toString()`, so the value must be an ISO date; `21-03-2026` was stored as a rule that never fires. Parsed with `LocalDate.parse` now.

Tightened in the model rather than only in the importer: the editor's date field is **free text** (labelled "Date (yyyy-MM-dd)"), so it could produce the same broken value — its save-time validation now catches it too.

Verified on device that this doesn't regress the editor: built a rule with a `Date = 2026-03-21` condition and it saved and listed normally.

Full app suite green — codec 17/17, `RuleConditionTest` 21/21, `RuleEngineTest` 19/19, no failures anywhere.

sarim2000:
Round 5b — took this one, in `70b1e57a`. You were right, and this time there *was* an authority to enforce, which is why I did it after declining the earlier "action/field pairing" framing.

`RuleEngine.applyAction` implements a specific set of pairs; everything else falls to its `else`, leaves the transaction untouched and records no `FieldModification` — the rule is silently dead. Import could carry such a pair and it was persisted and reported as a success.

`supportedActionTypes(field)` now sits alongside `RuleAction`, mirroring `applyAction`:

| Field | Executable actions |
|---|---|
| `CATEGORY` | `SET`, `CLEAR` |
| `MERCHANT`, `NARRATION` | `SET`, `APPEND`, `PREPEND`, `CLEAR` |
| `TYPE` | `SET` |
| `BANK_NAME` | `SET`, `CLEAR` |
| everything else | — |

`BLOCK` is exempt: it never modifies a transaction (the engine skips it in `applyActions` and handles it in `shouldBlockTransaction`), so it's valid on any field.

**The table can't drift from the engine.** `RuleActionExecutabilityTest` drives every field × action pair through `evaluateRules` and asserts the table agrees with whether the transaction actually changed. It passes, which is also the evidence the table is correct rather than my guess at it.

**One thing deliberately left alone:** `ADD_TAG` / `REMOVE_TAG` are executable *nowhere* in the app — they exist only as picker labels in `CreateRuleScreen`. The importer excludes them like any other non-executable action, but the real bug is that the editor still offers them, letting users build dead rules. That's pre-existing and outside this PR; I've flagged it separately for @sarim2000.

Codec tests 19/19, full app suite green.

sarim2000:
Round 6 — both real, fixed in `4da83c81`. The first turned out to be an engine bug, not an import-validation one.

**Amount equality never matched.** `EQUALS` / `NOT_EQUALS` compared `getFieldValue` — which is `BigDecimal.toString()`, so `"100.00"` — against the rule's raw text. A rule of `AMOUNT = 100` could never fire, and neither could `= 100.0`. The editor offers `=` for amounts, so this hit hand-built rules too, not just imported ones.

Those two operators now compare numerically when both sides parse, reusing the `compareNumeric` the ordering operators already used. Every other field keeps its plain text match. I fixed the engine rather than rejecting such rules on import, because refusing a numeric equality would be the wrong answer when the right one is for it to work.

**`TYPE` + `SET` accepted any non-blank value.** `applyAction` falls back to the transaction's existing type when the value won't parse, so `SET TYPE banana` silently did nothing. `RuleAction.validate()` now requires a parseable `TransactionType` — mirroring the check the TYPE *condition* already had.

Tests, all green with no failures anywhere in the suite:

| Suite | |
|---|---|
| `RuleEngineTest` | 23/23 (4 new: numeric equality matches, rejects a different number, `NOT_EQUALS` is numeric too, non-amount equality still textual) |
| `RuleSharingCodecTest` | 21/21 (2 new: bad TYPE value refused, valid one imports) |
| `RuleActionExecutabilityTest` | 2/2 — still agrees with the engine after the change |
| `RuleConditionTest` | 21/21 |

PR #758, opened 2026-09-07
fix(loans): only suggest people with an open loan, and show the note

### LINKED ISSUE #753: [BUG] Hide inactive /settled loan names
## 🐛 Bug Description
 Hide inactive /settled loan names when marking a transaction as loan and only display the active loan members

## 📱 Steps to Reproduce
1. Go to a transaction 
2. Click on mark as loan
4. See error you can see all the loans that you ever created
5. 
## ✅ Expected Behavior
Only loans that are not settled yet should be visible 

## 📸 Screenshots
I can see these many names whil I only have 4 loans that are unsettled
Image

## 📋 Device Information
- **Device**: moto g34 5g
- **Android Version**: Android 15
- **App Version**: 2.19.0

### LINKED ISSUE #754: [BUG] Loan notes are not displayed anywhere
## 🐛 Bug Description
The notes that we add while marking a transaction as loan are never displayed anywhere in the app

## 📱 Steps to Reproduce
1. Go to transaction
2. Click on mark as loan
3. Enter a note 
4. Open loans page
5. See error

## ✅ Expected Behavior
The loan note along with the transaction details should be visible in the loans page

## 📸 Screenshot

Image

## 📋 Device Information
- **Device**: moto g34 5g
- **Android Version**: Android 15
- **App Version**: 2.19.0

sarim2000:
Closes #753, closes #754. Two small loan fixes, 16 lines.

## #753 — settled people cluttered the picker

`getRecentPersonNames()` was `SELECT DISTINCT person_name FROM loans` with no status filter, so "mark as loan" suggested everyone you'd ever lent to. The reporter had 4 open loans and a screen full of names.

Added `WHERE status = 'ACTIVE'`. A settled person can still be typed in by hand — this only changes the suggestion chips.

## #754 — the note went nowhere

`LoanEntity.note` is captured in the "mark as loan" sheet, stored, and then rendered on no screen at all. It now appears under the amount on the loan detail, above the progress bar — it's the only place the user says what the loan was *for*, so it belongs with the figure rather than below the transaction history.

## Verified on the emulator

Lent ₹500 to "Alice" with the note "dinner", then settled it; lent ₹100 to "Bob" and left it open.

| | |
|---|---|
| Loan detail | note renders under the amount |
| Picker after settling Alice | **Alice absent** |
| Same picker | **Bob present** |

That last row is the one that matters — it shows the filter removed the right thing rather than breaking suggestions outright.


sarim2000:
On the non-blocking suggestion — I'm skipping the regression test, deliberately.

The change is one `WHERE status = 'ACTIVE'` in a Room `@Query` string. Covering it means executing that SQL, and the repo has no DAO test harness at all: zero `Room.inMemoryDatabaseBuilder` usages anywhere, and `androidTest` contains only the template `ExampleInstrumentedTest`. So the test isn't "add a case to the suite", it's "stand up the project's first instrumented database harness" — a lot of new scaffolding guarding one line.

What I did instead was verify the behaviour end to end on the emulator, including the control that a filter change actually needs:

| | |
|---|---|
| Alice — loan settled | **absent** from suggestions |
| Bob — loan still open | **present** |

The second row is the one that matters: it shows the filter removed the right thing rather than emptying the list, which a query-level test asserting "settled names are excluded" could pass while still being broken.

If DAO coverage is wanted, it's worth doing as its own piece of work with a shared harness that several DAOs can use — happy to pick that up separately.

Issue #749, opened 2026-09-03
[BUG] expense split breaks analytics


rohan5:
## 📱 Steps to Reproduce
1. Make an expense that goes to shopping by default.
2. Now update the expense to split between other two categories except shopping

## ✅ Expected Behavior

Expense should be visible in their respective categories, but instead, it is being shown completely in original shopping category, as well as partially in one of the other category. When I click on the third category, it takes me to transactions now, instead of that specific category transactions.

## 📋 Device Information
- **Device**: 
- **Android Version**: 
- **App Version**:

sarim2000:
Thanks — the second part is fixed in #768: tapping a split category from Analytics now opens the Transactions tab correctly filtered instead of showing everything. Root cause was the filter only matching a transaction's *parent* category, so a category that existed only inside a split matched nothing and the filter was auto-cleared.

The first part — *"shown completely in the original category, as well as partially in one of the other"* — I couldn't reproduce. On a fresh split (Others → Groceries 450 + Transportation 450) the Analytics breakdown reads Groceries ₹450 (50%) + Transportation ₹450 (50%), total ₹900, with the original category not listed at all. Two things would help pin it down:

1. **Which screen** shows the double-count — Analytics → Top Categories, the Budgets screen, or the Home "Spent this month" card?
2. Did you split the transaction **once**, or edit an existing split afterwards (e.g. change a category or amount on a split you'd already saved)?

If you can also say the three category names involved and whether the total at the top matches the sum of the rows, that'd narrow it fast.

sarim2000:
Closing the remaining half as not reproducible.

The part I could reproduce is fixed and shipped: tapping a split category in Analytics now opens Transactions filtered correctly, instead of showing everything.

The double-count — a split showing fully under its original category and again under one of the new ones — I couldn't reproduce on a fresh split, and the follow-up questions went unanswered. If you still see it, reopen with which screen shows it (Analytics, Budgets or the Home card) and whether the split was saved once or edited afterwards.

PR #768, opened 2026-09-13
fix(transactions): make the single-category filter split-aware (#749)

sarim2000:
Part of #749 — fixes the second symptom, and reports on the first.

## Fixed: tapping a split category showed everything

Tapping a category in Analytics opens the Transactions tab filtered on `transaction.category` **alone**. A transaction split into Groceries + Transportation keeps its parent category (by design), so a split-only category matched nothing — and the available-categories auto-clear, which only knew parent categories, then dropped the filter and showed **all** transactions. That's the reporter's *"takes me to transactions instead of that specific category"*.

The budget filter a few lines down was already split-aware, so the single category now routes through it as the one-element case rather than its own parent-only query, and the available-categories list includes split categories so the auto-clear no longer discards a filter that is valid.

Verified on the emulator with a split transaction (Others → Groceries 450 + Transportation 450) among unrelated rows in other categories:

| tap "Transportation" in Analytics | rows shown |
|---|---|
| before | every category — Entertainment, Healthcare, Mobile, Shopping… |
| after | the split row + the one genuine Transportation row, chip still set |

## Not reproduced: the double-count

I could not reproduce *"shown completely in the original category, as well as partially in one of the other"* in Analytics on a fresh split. `getAmountByCategory()` uses only the split amounts when splits exist, and on the emulator the breakdown read Groceries ₹450 (50%) + Transportation ₹450 (50%), total ₹900, no "Others" — correct. I've asked the reporter on the issue which screen shows the double-count and whether the split was edited after creation; leaving #749 open for that.

`./init.sh app` green.


sarim2000:
Fair point, fixed in the follow-up commit.

The split lookup was a one-shot `suspend` query inside a reactive flow, so a split edited while the list was open could leave the filter — and the category chips — stale until the transactions flow happened to re-emit for another reason. Both call sites now observe the splits through a `Flow` form of the same DAO query (`observeSplitsForTransactions`), so a split edit re-evaluates them directly.

Net effect on the code is a simplification: the hand-rolled `flow { … emit(filtered) }` builder becomes a `map` over the observed splits.

`./init.sh app` green.

PR #769, opened 2026-09-13
feat(rules): make Add Tag / Remove Tag actions actually apply (#748)

### LINKED ISSUE #748: [FEATURE] Smart Rules: Action to add Tags
## 💡 Feature Description

In Smart Rules THEN part, an action `Set Tag`, or even better, `Append Tag` would be very useful.

## 🎯 Problem it Solves

When I have transactions which belong to the Other category, I still want to be able to tag it, for example, to the platform the purchase belongs to, for example Swiggy.

## 🔧 Proposed Solution

In the Action Drop down, an Append Tag action should be available

## 📱 Alternatives Considered

Maybe changing the merchant works for this specific use case, but in general ability to modify tags through rules seems super useful!

## 💬 Additional Context

love you

sarim2000:
Closes #748.

## The gap

`ADD_TAG` / `REMOVE_TAG` existed in the enum and the rule editor offered them, but nothing applied them: `RuleEngine.applyAction` returns a modified `TransactionEntity`, and tags live in a separate table keyed by transaction id. A user could build a tag rule, see it listed as active, and get no effect ever. #746 documented this and pinned it with `RuleActionExecutabilityTest`.

## Design

- `TransactionField.TAGS`, **action-only** — `supportedOperators(TAGS)` is empty, so the importer refuses a TAGS *condition* (the pure engine has no tag value to compare).
- The engine stays pure. Tag actions never touch the entity; they're recorded as `FieldModification(TAGS, …)` on the `RuleApplication` the engine already returns.
- One shared helper, `List<RuleApplication>.applyTagActions(existing)`: merges adds case-insensitively without duplicates, removes case-insensitively, and returns `existing` by identity when nothing changed so callers skip the write.
- Applied at **all three** persist sites after the row id is known — SMS worker, SMS receiver, and apply-to-past — so it works on old transactions, not just new SMS. `ADD_TAG` merges; it never replaces.
- Editor: "Tags" as an action field offering Add/Remove Tag with a free-text name; switching fields coerces the type so a dead Set-on-Tags can't be saved. Dry-run preview shows `Tags: +Swiggy`.

## Two bugs the device check found

**Pre-existing on `main`:** the SMS receiver path saved rule-application audit rows straight from the engine, where the transaction id is still the pre-insert placeholder `0`. The FK rejected the write and the throw skipped everything after it — including the **balance update** — for *any* rule matching a live SMS. Tag rules just made it visible. Now remapped to the row id, as the worker already did.

**Preview overstated:** it listed the rule's intent, so a row that already carried the tag was still "would be updated". It now diffs against existing tags (read once for all rows).

## Verified on the emulator, against the database

| step | result |
|---|---|
| Add Tag rule, preview | `Tags: +TestTag`, 1 would be updated |
| Apply to past | tag on the matching row **only** — 1 link across 45 transactions |
| Incoming SMS matching the rule | `Applied 1 rules` → `Saved new transaction` → `Saved balance update`, no FK error; row tagged, `rule_applications` keyed correctly |
| Preview again | reports only the untagged match, not the already-tagged ones |

`./init.sh app` green — 303 tests, 0 failures. `RuleActionExecutabilityTest` now defines "carried out" as *entity changed **or** a TAGS modification was recorded*; every other field/action pair is asserted exactly as before.


sarim2000:
Both findings were right — fixed in the latest commit.

**P1 — concurrent edits lost.** All three persist sites did read-all → merge → replace-all, so a manual tag edit between the read and the write was overwritten with the rule's stale snapshot. Replaced with `TagRepository.applyTagChanges(id, add, remove)`: it inserts and deletes only the cross-refs the rule *names* (`IGNORE` on an existing pair), inside one `withTransaction`. It never touches a tag it wasn't told about, so there's nothing to clobber. Each call site is now two lines.

**P2 — no-ops counted as updates.** That same method returns whether anything actually changed, and the apply path now increments only when a field or a tag really moved — writing the row only when a field did. Completion and preview now agree.

Verified on the emulator, against the database:

| step | result |
|---|---|
| put a manual tag `Manual` on an untagged matching row | `Manual` |
| run the Add Tag rule | row now `Manual` **and** `TestTag` — the manual tag survived |
| completion dialog | "Transactions updated: 1" — only that row; the two already-tagged rows weren't counted |
| run it again | "No transactions match this rule" — nothing to change, nothing counted |

`./init.sh app` green; `RuleActionExecutabilityTest` at 6 with a new case for the add/remove extraction (later action wins, case-insensitive, blanks dropped).

Issue #374, opened 2026-05-29
[FEATURE] add sub categories


shreyes-shalgar:
## 💡 Feature Description
Need sub categories to have granular categorization

## 🎯 Problem it Solves
Categories more finer

## 🎨 UI/UX Suggestions (Optional)
Refer categorization in cashiro

shreyes-shalgar:
Is this feature being planned soon?
This is already beautifully implemented in cashiro. 
Im not sure but i think it was implemeted in cacf114e818e42f0f1e1f6844c9aec2c56c2e070 this commit

Pujari-Praveen:
Waiting on this feature

PR #780, opened 2026-09-14
fix(analytics): label the row with the sub-category when filtering on it (#374 follow-up)

sarim2000:
With the category filter set to a sub-category (e.g. *Coffee*), the Top Categories row was still labelled with the parent (*Food & Dining*) because the rollup applied unconditionally. Now the rollup is skipped only when the active filter is itself a sub-category, so the row reads *Coffee*. Every other view keeps rolling children into their parent.

Verified on emulator: Coffee filter → Top Categories shows a single *Coffee* row (₹55); parent filter and unfiltered views unchanged.


sarim2000:
Fixed the split case: only the filtered sub-category is exempt from rollup now (`parentOf - filter`), so the Groceries half of a Coffee/Groceries split still rolls into Food & Dining.

On the test ask: the branch is now a single `Map.minus(key)`; a unit test for that is asserting the stdlib, and an AnalyticsViewModel-level test needs a Hilt/Room harness the repo doesn't have (same call as #758). Skipping.

PR #781, opened 2026-09-14
feat(transactions): allow splits on credit-card purchases (#750)

### LINKED ISSUE #750: [FEATURE] expense split should also be available for credit transaction
## 💡 Feature Description
Expense split should also be available for credit transaction

sarim2000:
Closes #750

`CREDIT` (credit-card purchase) is a spend like `EXPENSE`, and every split consumer — analytics, budgets, the Transactions category filter, widgets — already reads splits per transaction regardless of type. So this only opens the two entry gates: `enableSplitMode()` and the *Split into categories* button now accept `EXPENSE` and `CREDIT` (`SPLITTABLE_TYPES`). Income/transfer/investment stay unsplittable.

Verified on emulator: switched a transaction to Credit → split button shown → split Coffee/Others 27.50 each → saved, both rows in `transaction_splits`.


sarim2000:
Widget: the category pie now includes CREDIT transactions (and therefore their splits) when the user's *Count card spend as expense* switch is on — the same rule Home uses. With it off, card purchases stay out of the pie as before, split or not.

